PR #35, opened 2016-11-19
Now depend on stratisd-client-dbus.

mulkieran:
Remove obsoleted tests and dependencies.


mulkieran:
@agrover review?

This assumes that stratisd PR for AddCacheDevs goes in.

agrover:
Thanks for letting me take a look -- it's a chance for me to start increasing my familiarity with this codebase.

Only Q I have: the first patch adds some sleep()s, I guess to allow the service to start up fully, export its bus name, etc... was this previously causing repeatable, or transient failures? Just curious. 👍

mulkieran:
The answer is a bit complicated. I've been working a while on stratisd-client-dbus tests were are a bit closer to the D-Bus than these CLI tests. The stratisd-client-dbus tests were failing all around because of calling methods before the service was set up, so I added sleeps uniformly in the setUp() methods, right after the service gets started and before any methods get called. The intention is to never add sleeps anywhere else, but this location seemed legitimate to me.

When I went back to working on the CLI tests, I imported a number of improvements to the tests from the stratisd-client-dbus test code and this was one of them. Failures on the CLI tests were quite infrequent, probably because starting up a whole Python process to talk to the D-Bus takes enough time in itself that the sleep generally isn't necessary. On the other hand, test failures due to those kinds of problems are annoying, and handling them in a piecemeal fashion is tedious, so I thought it would be better just to head off the problem before it could cause any distractions.

mulkieran:
I had to update to get rid of some stray imports that were no longer needed, but that's the only change.

PR #46, opened 2017-02-28
Some proposed changes to stratis command line help text

agrover:
More to come, but here are some initial proposed changes for cmdline help text, and to rename the `stratisd` subcommand to `daemon`.

### DIFF
--- a/src/stratis_cli/_actions/_physical.py
+++ b/src/stratis_cli/_actions/_physical.py
@@ -50,7 +50,7 @@ class PhysicalActions(object):
         """
         proxy = get_object(TOP_OBJECT)
         pool_object = get_object(
-           GetObjectPath.get_pool(proxy, spec={'Name': namespace.name})
+           GetObjectPath.get_pool(proxy, spec={'Name': namespace.pool_name})
         )
 
         (_, rc, message) = Pool.AddDevs(
--- a/src/stratis_cli/_parser/_parser.py
+++ b/src/stratis_cli/_parser/_parser.py
@@ -28,7 +28,7 @@ from ._physical import build_physical_parser
 from ._pool import build_pool_parser
 
 
-def build_stratisd_parser(parser):
+def build_daemon_parser(parser):
     """
     Generates the parser appropriate for obtaining information about stratisd.
 
@@ -58,7 +58,7 @@ _SUBPARSER_TABLE = {
    'blockdev' : build_physical_parser,
    'filesystem' : build_logical_parser,
    'pool' : build_pool_parser,
-   'stratisd' : build_stratisd_parser
+   'daemon' : build_daemon_parser
 }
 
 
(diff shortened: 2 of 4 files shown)


mulkieran:
These all look fine to me so far.

mulkieran:
Errr, the changes look fine, but it would be nice also to fix the tests. These are tests that are not run as part of the CLI. To run the tests, strart the daemon with '--sim', edit ```tests/constants``` appropriately, and then ```PYTHONPATH=<the stuff> make dbus-tests```.

agrover:
@mulkieran can you take a look at last two commits, especially? Also, tests are failing and I can't figure out why -- it looks like `name = arg['name']` in `add_args()` is failing but that should never happen (each args dict should always have a name)

README.rst
```diff
@@ -20,6 +20,10 @@ To install, check out the source, and use the included setup script, as::
 
    > python setup.py install
 
+You will also need to clone and install the following related Stratis
```
mulkieran: They can use pip as well, but maybe that should be a secret.

README.rst
```diff
@@ -48,23 +49,14 @@ Testing
...
-   > py.test tests
-
-at the command prompt. Entering::
-
-   > tox
-
-will cause some tests to be run in the tox virtual environment. Integration
-tests, which requires the CLI to connect with the Stratis daemon, will not
-be run.
+of your stratisd executable, for example::
```
mulkieran: The only tests tox still runs are lints, which are really more checks. But I wouldn't choose to stop mentioning it entirely.

README.rst
```diff
@@ -33,10 +37,7 @@ or::
 To run without installing, check out the source, change to the top directory
 and enter::
 
-   > ./bin/stratis --help
-
-making sure that your PYTHONPATH environment variable is set to the CLI src
-directory.
+   > PYTHONPATH="src:../stratisd-client-dbus/src:../into-dbus-python/src:../dbus-signature-pyparsing/src" python3 bin/stratis --help
```
mulkieran: How you set your environment variable depends on the shell you're using...

src/stratis_cli/_parser/_cache.py
```diff
@@ -13,123 +13,59 @@
...
-
-    subparser_table['remove'] = \
-       subparsers.add_parser(
-          'remove',
-          description="Remove Cache Device from a Pool Cache"
-       )
-    _SUBPARSER_TABLE['remove'](subparser_table['remove'])
-
-    return parser
+cache_subcmds = [
```
mulkieran: If these dicts were actually constants, i.e., if the use of ```del``` was removed, then they might as well be capitalized.

agrover:
I've dropped the last two commits -- will open separate PRs once ready -- from this PR and have updated the original commits to pass tests.

I'm going to go ahead and merge this now, hope that's ok.

Issue #44, opened 2017-02-28
Stratisd --redundancy and --version should be subcommands, not arguments


agrover:
Or even better, have `stratis stratisd` get and output both.

mulkieran:
I don't think that's better. Then I have to look at both, even if I only wanted one. And then I as the developer have to start worrying about formatting...and in a CLI that's just annoying.

Turning them from options to subcommands is fine, it makes very little difference.

PR #50, opened 2017-03-01
Miscellaneous changes

mulkieran:


agrover:
looks good. Can we do without --propagate? Is this for developer/debugging use or would a user actually find it useful?

agrover:
Also, you may choose to put `closes #44` in the commit log for fe30824 and it will automatically close that when this PR is merged; see https://help.github.com/articles/closing-issues-via-commit-messages/

mulkieran:
```---propagate``` is for developer use. Perhaps an environment variable would be better than a flag? Or maybe it should go under a subcommand, like ```developer```?

PR #51, opened 2017-03-02
Make tests portable

mulkieran:


mulkieran:
I had better update the README as well.

mulkieran:
It should be good to go now.

PR #56, opened 2017-03-10
Define commands using a data structure

agrover:
Incorporates review comments from #46.

Passes tests!


src/stratis_cli/_actions/_top.py
```diff
@@ -49,10 +49,10 @@ def create_pool(namespace):
 
         (_, rc, message) = Manager.CreatePool(
            proxy,
-           name=namespace.name,
+           name=namespace.pool_name,
            redundancy=0,
            force=namespace.force,
-           devices=namespace.device
+           devices=namespace.device_name
```
mulkieran: This variable name change seems really dubious to me. I don't think it really is the name. It could be a variety of things.

agrover: ok, changing to `blockdevs`, if that's better.

agrover:
@mulkieran I think that worked out well! please review/merge

src/stratis_cli/_parser/_parser.py
```diff
@@ -23,44 +23,81 @@
...
 
+def add_subcommand(subparser, cmd):
+    """
+    Add subcommand to a parser based on a subcommand dict.
+    """
+    name, info = cmd
+    parser = subparser.add_parser(name, help=info['help'])
+
+    subcmds = info.get('subcmds')
+    if subcmds:
```
mulkieran: Please do explicit comparison with None, as ```if subcmds is not None:``` and same on line 54. It should never make any difference here, but I've seen overloading in Python leading to an enormous number of bugs, so I'm only in favor of it when it is needeed.

src/stratis_cli/_parser/_parser.py
```diff
@@ -70,50 +107,22 @@ def gen_parser():
     :rtype: ArgumentParser
     """
     parser = argparse.ArgumentParser(
-       description="Stratis Storage Manager",
-       prog='stratis'
+        description="Stratis Storage Manager",
```
mulkieran: This through line 121 breaks the indenting policy of this library...4 spaces if semantic indentation, 3 for non-semantic.

src/stratis_cli/_parser/_parser.py
```diff
@@ -23,44 +23,81 @@
...
+         subcmds=LOGICAL_SUBCMDS,
+     )),
+    ('daemon',
+     dict(
+         help="Stratis daemon information",
+         subcmds=DAEMON_SUBCMDS,
+     )),
+]
+
+gen_args = [
```
mulkieran: Probably this should be capitalized as well.

src/stratis_cli/_parser/_parser.py
agrover: fixed

src/stratis_cli/_parser/_parser.py
agrover: fixed

src/stratis_cli/_parser/_parser.py
agrover: fixed, added to readme. I'm assuming what pep8 calls continuation lines are what you're referring to as non-semantic.

mulkieran: I hope so...

mulkieran:
Exercising my right to merge!

PR #58, opened 2017-03-18
List pool and filesystem names instead of object paths

agrover:
The given name is more useful to users than arbitrary object path.

### DIFF
--- a/src/stratis_cli/_actions/_logical.py
+++ b/src/stratis_cli/_actions/_logical.py
@@ -23,6 +23,7 @@ from stratisd_client_dbus import Pool
 from stratisd_client_dbus import StratisdErrorsGen
 from stratisd_client_dbus import get_managed_objects
 from stratisd_client_dbus import get_object
+from stratisd_client_dbus import GMOFilesystem
 
 from .._errors import StratisCliRuntimeError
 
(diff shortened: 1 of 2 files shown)


agrover:
@mulkieran I just used `info` as a placeholder variable name here since I couldn't think of a better one, I'm hoping you have a recommendation.

src/stratis_cli/_actions/_logical.py
```diff
@@ -66,10 +66,10 @@ def list_volumes(namespace):
         pool_object_path = \
            GetObjectPath.get_pool(proxy, spec={'Name': namespace.pool})
 
-        for object_path, _ in get_managed_objects(proxy).filesystems(
+        for _, info in get_managed_objects(proxy).filesystems(
                    props={'Pool': pool_object_path}
            ):
-            print(object_path)
+            print(info['org.storage.stratis1.filesystem']['Name'])
```
mulkieran: ```stratisd_client_dbus``` package defines two classes, ```GMOFilesystem``` and ```GMOPool```. The idea is that they get at the table, like: 

```print(GMOFilesystem(info).Name())```

which is a bit more principled. The same should work below, with ```GMOPool(info).Name()```. These should continue to exist once the whole thing gets generalized, so they might as well be used.

mulkieran:
I have no recommendation better than info :)

PR #67, opened 2017-08-17
Update README.rst

stardude900:
My default CentOS 7.x install includes installing Python3.6 (or whatever the latest released version is) and python-dbus has yet to be built for python 3.6

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -29,6 +29,25 @@ stratisd-client-dbus, into-dbus-python, dbus-signature-pyparsing.
 
 Finally, ensure you have python3-pyparsing package installed.
 
+Troubleshooting
+---------------
+Error::
+
+> Traceback (most recent call last):
+>   File "/usr/bin/stratis", line 4, in <module>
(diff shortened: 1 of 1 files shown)


agrover:
OK let me see if I can reproduce this issue. I'm running Centos 7.3.1611, which seems to be the latest.. I don't see any python3 until I add epel, and then all Python3 epel packages appear to be for python34.

stardude900:
Hm, I've tried on a VM at home (i've been out of the office since about 2
hours before your message) and can't duplicate it.  I'm wondering if the
very limited kickstart that I used for that VM did it.  I'll update again
when I've had a chance to work on it, probably tomorrow or Monday.

Regards,
Logan

stardude900:
What happened is the kickstart I used to create the VM installs the IUS
repo and then installs the latest python, which is 3.6.2.

So the entire incident was my mistake.

Regards,
Logan

agrover:
No problem! If you do try again with python34 from EPEL let me know how it goes and if you run into any issues. Closing this one.

PR #73, opened 2017-10-27
add snapshot command

trgill:


trgill:
I think the first two commits are correct - any feedback welcome.

I still need to deal with error handling.  For example - when the filesystem doesn't exist - we should print an error message.  

# bin/stratis filesystem snapshot pool nofs snapname
Execution failed: No object path found for interface org.storage.stratis1.filesystem and spec {'Name': 'nofs', 'Pool': dbus.ObjectPath('/org/storage/stratis1/1')}

src/stratis_cli/_actions/_data.py
```diff
@@ -91,6 +91,13 @@
 <arg name="return_code" type="q" direction="out"/>
 <arg name="return_string" type="s" direction="out"/>
 </method>
+<method name="SnapshotFilesystem">
```
mulkieran: Just want to make sure that this update got into the file by:
1. Firing up stratisd, creating a filesystem, and calling Introspect() on the filesystem object.
2. Overwriting whatever was previously in this file for Filesystem w/ the new value.

Otherwise, future editors may run into unnecessary conflicts.

trgill: @mulkieran I called Instrospect on the pool (not filesystem) via:

busctl call org.storage.stratis1 /org/storage/stratis1/1 org.freedesktop.DBus.Introspectable Introspect

But the format didn't match what was in the code.  I made some format adjustments by hand.  Is there a formatting flag you passed busctl?

mulkieran: I didn't use busctl directly, I used one of the GUI wrappers, w/out realizing it wasn't doing the same thing as busctl. Maybe we should have a policy. But in any case, since the order is exactly that which stratisd generates, that is good enough for now.

trgill: Ok.  I looked at it with d-feet and it is similar - but would have also required some adjustments to match the code.

mulkieran: Yeah. I used the other GUI, I forget it's name right now. It's the one @agrover recommended. Hmm.

trgill:
@mulkieran Ok.  Updated to use the _MENU approach like in test_create.  I realize that Create2TestCase is skipped.  But, is the _MENU variable correct in that function?  It shouldn't have the 'pool' entry - right?  I'm happy to fix it - just making sure I'm not missing something.

mulkieran:
@trgill For that test...it probably makes sense to fix it up and stop skipping it in a separate PR. I think the only reason it's skipped is that there could be some possibility that the pool _might_ be out of space. But since this is a basic exercise of the D-Bus I think it makes sense to restore it.

tests/integration/logical/test_snapshot.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        Stop the stratisd simulator and daemon.
+        """
+        self._service.tearDown()
+
+    def testCreation(self):
+        """
+        Creation of the snapshot must fail since filesystem does not exist.
+        """
+        command_line = ['pool', 'snapshot'] + [self._POOLNAME] + self._FSNAME + self._SNAPNAME
```
mulkieran: This one should also use ```_MENU```.

mulkieran:
You will have to find out why Jenkins doesn't seem to be running the tests in test_snapshot.

tests/integration/logical/test_snapshot.py
```diff
@@ -0,0 +1,133 @@
...
+    _FSNAME = 'fs'
+
+    def setUp(self):
+        """
+        Start the stratisd daemon with the simulator.
+        """
+        self._service = Service()
+        self._service.setUp()
+        time.sleep(1)
+        command_line = \
```
mulkieran: There's probably some oversights in some examples, but better to be less convoluted with lists, like:

```['pool', 'create', self._POOLNAME, _DEVICE_STRATEGY.example()]```

File: tests/integration/logical/test_snapshot.py [hunk 4]
```diff
@@ -0,0 +1,133 @@
...
+        """
+        Stop the stratisd simulator and daemon.
+        """
+        self._service.tearDown()
+
+    def testSnapshot(self):
+        """
+        Creation of the snapshot should succeed since origin pool/filesytem is available.
+        """
+        command_line = self._MENU + [self._POOLNAME] + [self._FSNAME] + [self._SNAPNAME]
```
mulkieran: Here, ```self._MENU + [self._POOLNAME, self._FSNAME, self._SNAPNAME]``` is the way to go.

tests/integration/logical/test_snapshot.py
```diff
@@ -0,0 +1,133 @@
...
+        """
+        Stop the stratisd simulator and daemon.
+        """
+        self._service.tearDown()
+
+    def testCreation(self):
+        """
+        Creation of the snapshot must fail since pool is not specified.
+        """
+        command_line = self._MENU + [self._POOLNAME] + [self._FSNAME] + [self._SNAPNAME]
```
mulkieran: This line can be tightened up.

tests/integration/logical/test_snapshot.py
```diff
@@ -0,0 +1,133 @@
...
+    _SNAPNAME = 'snapfs'
+
+    def setUp(self):
+        """
+        Start the stratisd daemon with the simulator.
+        """
+        self._service = Service()
+        self._service.setUp()
+        time.sleep(1)
+        command_line = \
```
mulkieran: This line can be tightened up.

tests/integration/logical/test_snapshot.py
```diff
@@ -0,0 +1,133 @@
...
+        """
+        Stop the stratisd simulator and daemon.
+        """
+        self._service.tearDown()
+
+    def testCreation(self):
+        """
+        Creation of the snapshot must fail since filesystem does not exist.
+        """
+        command_line = self._MENU + [self._POOLNAME] + [self._FSNAME] + [self._SNAPNAME]
```
mulkieran: This line can be tightened up.

tests/integration/logical/test_snapshot.py
```diff
@@ -0,0 +1,133 @@
...
+        Start the stratisd daemon with the simulator.
+        """
+        self._service = Service()
+        self._service.setUp()
+        time.sleep(1)
+        command_line = \
+           ['pool', 'create'] + [self._POOLNAME] + \
+           _DEVICE_STRATEGY.example()
+        RUNNER(command_line)
+        command_line = ['filesystem', 'create'] + [self._POOLNAME] + [self._FSNAME]
```
mulkieran: Simplify list here as well.

tests/integration/logical/test_snapshot.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def tearDown(self):
+        """
+        Stop the stratisd simulator and daemon.
+        """
+        self._service.tearDown()
+
+    def testCreation(self):
+        """
+        Creation of the snapshot must fail since pool is not specified.
```
mulkieran: "is not specified" -> "specified pool does not exist".

tests/integration/logical/test_snapshot.py
trgill: ok

tests/integration/logical/test_snapshot.py
trgill: ok

tests/integration/logical/test_snapshot.py
trgill: ok

tests/integration/logical/test_snapshot.py
trgill: ok

tests/integration/logical/test_snapshot.py
```diff
@@ -0,0 +1,131 @@
...
+    _SNAPNAME = 'snapfs'
+
+    def setUp(self):
+        """
+        Start the stratisd daemon with the simulator.
+        """
+        self._service = Service()
+        self._service.setUp()
+        time.sleep(1)
+        command_line = ['pool', 'create'] + [self._POOLNAME] + _DEVICE_STRATEGY.example()
```
mulkieran: I think this must be able to be combined into:

```['pool',  'create', self._POOLNAME, _DEVICE_STRATEGY.example()]``` or something very close.

File: tests/integration/logical/test_snapshot.py [hunk 11]
```diff
@@ -0,0 +1,131 @@
...
+
+    def setUp(self):
+        """
+        Start the stratisd daemon with the simulator.
+        """
+        self._service = Service()
+        self._service.setUp()
+        time.sleep(1)
+        command_line = \
+           ['pool', 'create'] + [self._POOLNAME] + \
```
mulkieran: This one should be able to be tightend up to a single list, no concat, ```+```, operators anywhere.

tests/integration/logical/test_snapshot.py
```diff
@@ -0,0 +1,131 @@
...
+from .._misc import Service
+
+
+_DEVICE_STRATEGY = _device_list(1)
+
+
+class SnapshotTestCase(unittest.TestCase):
+    """
+    Test creating a snapshot of a filesystem in a pool.  In this case
+    the snapshot should be created and no error rasied.
```
mulkieran: Typo here, "rasied".

tests/integration/logical/test_snapshot.py
trgill: ok - I followed an example in test_create.py - should I add another commit to update the way the commands are constructed in that file too?

mulkieran: Probably not worth the trouble for existing code.

trgill: Ok.  Most (all?) of the other test cases don't follow this convention.  But, I made the updates.

tests/integration/logical/test_snapshot.py
trgill: ok

tests/integration/logical/test_snapshot.py
mulkieran: yeah. Bitrot. If it seems like it'll give people bad habits, feel free to fix it.

PR #75, opened 2017-10-30
Use justbytes lib to print values with units

### LINKED ISSUE #74: Sizes should be expressed in human units
e.g. in "stratis pool list", using a function like:
```
def sizeof_fmt(num, suffix='B'):
    for unit in ['','Ki','Mi','Gi','Ti','Pi','Ei','Zi']:
        if abs(num) < 1024.0:
            return "%3.1f%s%s" % (num, unit, suffix)
        num /= 1024.0
    return "%.1f%s%s" % (num, 'Yi', suffix)
```
(from https://stackoverflow.com/questions/1094841/reusable-library-to-get-human-readable-version-of-file-size)

agrover:
We only do this in one spot now but will likely be printing such values
much more in the future.

fixes #74

### DIFF
--- a/src/stratis_cli/_actions/_top.py
+++ b/src/stratis_cli/_actions/_top.py
@@ -18,12 +18,15 @@ Miscellaneous top-level actions.
 
 from __future__ import print_function
 
+from justbytes import Range
+
 from .._errors import StratisCliRuntimeError
 
 from .._stratisd_constants import StratisdErrors
 
 from ._connection import get_object
 from ._constants import TOP_OBJECT
+from ._constants import SECTOR_SIZE
 from ._data import Manager
 from ._data import MOPool
 from ._data import ObjectManager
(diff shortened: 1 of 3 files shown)


src/stratis_cli/_actions/_top.py
```diff
@@ -75,8 +78,8 @@ def list_pools(_):
         tables = [
            [
               mopool.Name(),
-              mopool.TotalPhysicalSize(),
-              mopool.TotalPhysicalUsed()
+              str(Range(mopool.TotalPhysicalSize(), SECTOR_SIZE)),
+              str(Range(mopool.TotalPhysicalUsed(), SECTOR_SIZE)),
```
mulkieran: You need to fix heading strings, below.

src/stratis_cli/_actions/_top.py
```diff
@@ -75,8 +78,8 @@ def list_pools(_):
         tables = [
            [
               mopool.Name(),
-              mopool.TotalPhysicalSize(),
-              mopool.TotalPhysicalUsed()
+              str(Range(mopool.TotalPhysicalSize(), SECTOR_SIZE)),
```
mulkieran: Note that you're accepting the default configuration here. The defaults were chosen by me as the most useful for humans representation I could come up with. But I'm not everybody's kind of human.

agrover: heh. Well it looked ok to me!

src/stratis_cli/_actions/_top.py
agrover: ok.

agrover:
made changes.

agrover:
ok, justbytes>=0.10 in setup.py.

PR #77, opened 2017-10-30
Initial Blockdev support

agrover:
Add introspection schema, print info on a pool's blockdevs, and fix a format error.

Waiting on PR #75 to be merged before using `justbytes` here, and then we'll be updating the last commit in this PR to also use it for printing size.

### DIFF
--- a/src/stratis_cli/_actions/_physical.py
+++ b/src/stratis_cli/_actions/_physical.py
@@ -18,16 +18,32 @@ Miscellaneous physical actions.
 
 from __future__ import print_function
 
+from justbytes import Range
+
 from .._errors import StratisCliRuntimeError
-from .._errors import StratisCliUnimplementedError
 from .._stratisd_constants import StratisdErrors
 
 from ._connection import get_object
+from ._constants import SECTOR_SIZE
 from ._constants import TOP_OBJECT
+from ._constants import UNKNOWN_VALUE_MARKER
+from ._data import devs
+from ._data import MODev
 from ._data import ObjectManager
 from ._data import Pool
 from ._data import pools
+from ._formatting import print_table
 
+
+def state_val_to_string(val):
+    """
+    Convert a blockdev state enumerated value to a string.
+    """
+    states = ["Missing", "Bad", "Spare", "Not-in-use", "In-use"]
+    try:
+        return states[val]
+    except IndexError:
+        return UNKNOWN_VALUE_MARKER
 
 class PhysicalActions(object):
     """
@@ -39,8 +55,35 @@ class PhysicalActions(object):
         """
         List devices in a pool.
         """
-        # pylint: disable=unused-argument
-        raise StratisCliUnimplementedError("No ability to list pools.")
+        proxy = get_object(TOP_OBJECT)
+        managed_objects = ObjectManager.Methods.GetManagedObjects(proxy, {})
+        (parent_pool_object_path, _) = pools(
+           managed_objects,
+           props={'Name': namespace.pool_name},
+           unique=True
+        )
+        modevs = [MODev(info) for _, info in devs(
+            managed_objects,
+            props={"Pool": parent_pool_object_path},
+        )]
+        tables = [
+            [
+                modev.Devnode(),
+                str(Range(modev.TotalPhysicalSize(), SECTOR_SIZE)),
+                state_val_to_string(modev.State()),
+            ] for modev in modevs
+        ]
+        print_table(
+            [
+                "Device Node",
+                "Physical Size",
+                "State",
+            ],
+            tables,
+            ['<', '>', '>']
+        )
+
+        return
 
     @staticmethod
     def add_device(namespace):
--- a/docs/stratis.txt
+++ b/docs/stratis.txt
@@ -76,7 +76,13 @@ GitHub for issues and development::
 'stratisd' projects, based on likelihood of issue being with the command-line
 tool or the service daemon.
 Mailing list::
-	stratis-devel@lists.fedorahosted.org for general development discussion
+	stratis-devel@lists.fedorahosted.org for general development
+	discussion
+Unknown values::
+	stratis-cli uses the Stratis API. If the API returns values that
+	stratis-cli cannot interpret, stratis-cli will substitute "???". If
+	encountered, upgrading to the latest version of stratis-cli, or filing
+	an issue, is recommended.
 
 LICENSE
 -------
--- a/src/stratis_cli/_actions/_constants.py
+++ b/src/stratis_cli/_actions/_constants.py
@@ -27,3 +27,7 @@ REDUNDANCY = {
 }
 
 SECTOR_SIZE = 512
+
+# This string is used in all instances where an uninterpretable value is
+# retrieved from the API.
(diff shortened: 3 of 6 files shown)


src/stratis_cli/_actions/_physical.py
```diff
@@ -27,8 +27,21 @@
 from ._data import ObjectManager
 from ._data import Pool
 from ._data import pools
+from ._data import MODev
+from ._data import devs
+from ._formatting import print_table
 
 
+def state_num_to_string(val):
```
mulkieran: Should have a header.

src/stratis_cli/_actions/_physical.py
```diff
@@ -27,8 +27,21 @@
...
+    states = {
+        0: "Missing",
```
mulkieran: Better to use a list, since these are just contiguous numeric indices.

src/stratis_cli/_actions/_physical.py
```diff
@@ -27,8 +27,21 @@
...
+        1: "Bad",
+        2: "Spare",
+        3: "Not-in-use",
+        4: "In-use",
+    }
+    return states.get(val, "Unknown")
```
mulkieran: Better to raise exception on unknown value.

src/stratis_cli/_actions/_physical.py
```diff
@@ -39,8 +52,35 @@ def list_pool(namespace):
...
+        )
+        modevs = [MODev(info) for _, info in devs(
+            managed_objects,
+            props={"Pool": parent_pool_object_path},
+        )]
+        tables = [
+            [
+                modev.Devnode(),
+                modev.TotalPhysicalSize(),
+                state_num_to_string(modev.State()),
```
mulkieran: Better to print out something disturbing if state numeric code really is unexpected, something like:
```UNEXPECTED NUMBER: 32```

File: src/stratis_cli/_actions/_physical.py (hunk 1 above)
Comment (agrover) [reply src/stratis_cli/_actions/_physical.py]: ok

File: src/stratis_cli/_actions/_physical.py (hunk 2 above)
Comment (agrover) [reply src/stratis_cli/_actions/_physical.py]: ok

File: src/stratis_cli/_actions/_physical.py (hunk 3 above)
Comment (agrover) [reply src/stratis_cli/_actions/_physical.py]: I disagree, because of the context this value is used. It's just informational. We could raise an exception, but then in the caller we're just going to catch it and convert back into an "UNEXPECTED" string to print, why not just do that in the function?

**Comment by agrover** [comment]
* use justbytes
* use a list in `state_val_to_string`
* include value in string if unknown

one weird thing: this is the output I'm getting:
```
Device Node    Physical Size   State
/dev/vdb            1020 MiB  In-use
/dev/vdc          < 3.00 GiB  In-use
```

why is the "<" in there? Is this because of the default justbytes string representation, and is it configurable?

**Comment by mulkieran** [comment]
Yes, and yes. But it does have a meaning which I would consider valuable, which is that the actual value is some amount less than 3.00 GiB, but the representation would need more than two decimal points to show the difference.

**Comment by agrover** [comment]
ok thanks. Yeah we might want to change the default but I'll open a separate issue.

**Comment by agrover** [comment]
* updated to fix pylint errors

File: src/stratis_cli/_actions/_physical.py [hunk 5]
```diff
@@ -18,16 +18,32 @@
...
+from ._data import MODev
+from ._data import devs
+from ._formatting import print_table
+
 
+def state_val_to_string(val):
+    """
+    Convert a blockdev state enumerated value to a string.
+    """
+    states = ["Missing", "Bad", "Spare", "Not-in-use", "In-use"]
```
mulkieran: is it possible to, instead:

* Make ```states``` a module level constant.
* Omit this function entirely.
* Change table calculation as I suggest below.

?

src/stratis_cli/_actions/_physical.py
```diff
@@ -39,8 +55,35 @@ def list_pool(namespace):
...
+        )
+        modevs = [MODev(info) for _, info in devs(
+            managed_objects,
+            props={"Pool": parent_pool_object_path},
+        )]
+        tables = [
+            [
+                modev.Devnode(),
+                str(Range(modev.TotalPhysicalSize(), SECTOR_SIZE)),
+                state_val_to_string(modev.State()),
```
mulkieran:
```
try:
  tables = [
    [ ...,
      ...,
      STATES[modev.State()]
    ] for modev in modevs
  ]
except IndexError as err:
  raise StratisCliError("horrors, stratis cli astonished by stratisd value, try updating to a new version of stratis-cli and./or reporting a bug") from err
```

To encourage the user to update their version of stratis so it works w/ the daemon again and if there is no new version of stratis-cli to treat this as a bug.

agrover: I don't think that's the right approach for us, since we're systems software. We need to degrade gracefully. Since blockdev state is an informational field only, an out-of-range value is harmless, and should not preclude the user being able to list the blockdevs that make up the pool. We cannot assume that users are always in a position where they are able to do an update or file a bug report. (e.g. maybe on a plane?) but they still may need to use our software for some reason.

mulkieran: An out-of-range value for this bit of information is just a symptom of the fact that the CLI is _ getting information from the daemon that it does not understand_. Bad information leads to bad decision making, which can go anywhere. The user should be alerted to the problem. What we're withholding by raising an exception could be some very misleading data. We should not downplay this fact. However, I agree that an exception might be a bit aggressive. How about:

```
STATES[[modev.State()] if modev.State() < STATES.len() else <STANDARD FORMULA TO INDICATE THAT THE RESULT FROM THE DAEMON WAS UNRECOGNIZED>
```

The STANDARD FORMULA can be documented in the CLI manual, so a user would recognize what it means wherever they encounter it.

agrover: Ok, that works. What were you thinking? Something like `???` or `Unrecognized value`?

mulkieran: ```???``` is more unsettling, so probably better.

mulkieran:
Also, there are a bunch of skipped tests `tests/integration/physical/test_list.py` which should be amended, removed, or added to as appropriate.

agrover:
working on tests and "universal unknown string" thing.

agrover:
Rebased a little.
* Enabled tests in physical/test_list.py
* Took a stab at the unknown value sentinel string

src/stratis_cli/_actions/_physical.py
```diff
@@ -18,16 +18,32 @@
...
 from .._stratisd_constants import StratisdErrors
 
 from ._connection import get_object
 from ._constants import TOP_OBJECT
+from ._constants import SECTOR_SIZE
+from ._constants import UNKNOWN_VALUE_MARKER
 from ._data import ObjectManager
 from ._data import Pool
 from ._data import pools
+from ._data import MODev
```
mulkieran: Must alphabetize.

agrover:
Alphabetized. Merging once CI is done.

PR #84, opened 2017-11-02
Sort rows of tables when printing

agrover:
Natural sorting order seems to give good results.


mulkieran:
I think that the caller should be allowed to sort the lines, according to their taste.

agrover:
ok, callers must sort. Closing.

PR #88, opened 2017-11-06
Now catch dbus-python-client-gen runtime errors in main

mulkieran:
Previously the name was not exported, so they could not be caught
explicitly.


mulkieran:
@agrover Review please?

Jenkins is just broken now, so the failure is information-free.

PR #90, opened 2017-11-07
Update command_line construction to be consistent

trgill:
Use 80 chars where possible
Combine strings into the same array if possible

Signed-off-by: Todd Gill <tgill@redhat.com>


trgill:
@mulkieran ok

PR #95, opened 2017-12-08
Filesystem rename feature

### LINKED ISSUE #93: Implement filesystem rename
We should be able to rename filesystems from the commandline. This is possible using the Filesystem SetName method.

carzacc:
It works in my tests using Fedora 26 Server.
Closes #93


agrover:
Hi, thanks for tackling this! Looks good except `stratis filesystem rename` needs to take `pool_name` in addition to old and new names. We can have two pools, each with a filesystem named "foo", which will cause an error because we're expecting the code that looks up the filesystem path to return a unique result.

carzacc:
@agrover Is it fine now?

agrover:
Looks great now! Much thanks!

PR #96, opened 2017-12-09
Add fs as an alias of filesystem

### LINKED ISSUE #65: add 'fs' as an alias (or replacement) for 'filesystem' subcommand
Yes, tab-completion will help but why not just make the subcommand name shorter?

carzacc:
I had to use info.get() to avoid getting [KeyErrors](https://wiki.python.org/moin/KeyError) when parsing subcommands with no aliases.
Closes #65

### DIFF
--- a/src/stratis_cli/_parser/_parser.py
+++ b/src/stratis_cli/_parser/_parser.py
@@ -43,7 +43,7 @@ def add_subcommand(subparser, cmd):
     Add subcommand to a parser based on a subcommand dict.
     """
     name, info = cmd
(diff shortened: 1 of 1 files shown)


mulkieran:
This looks fine to me. @agrover ?

agrover:
I think this is good. I think the next question is, do we want 'fs' as an alias or just replace 'filesystem' with 'fs'? And, how to treat this alias in the manpage? But let's merge this to start. Thanks! Merging as last reviewer.

PR #97, opened 2017-12-12
Now depend on justbytes 0.11

mulkieran:
The only important change is that it restricts the allowed versions of
justbases. There are also some doc fixes.


mulkieran:
@agrover review, please?

PR #103, opened 2018-01-08
Added completion for block device names

carzacc:
I added autocompletion for block device names and also for the fs names for filesystem destroy when the user wants to destroy more than one filesystem

### DIFF
--- a/etc/bash_completion.d/stratis
+++ b/etc/bash_completion.d/stratis
@@ -15,7 +15,7 @@ _stratis()
 	fs_subcommands="create snapshot list rename destroy"
 	blockdev_subcommands="add list"
 	daemon_subcommands="redundancy version"
-	blockdevs=$(echo /dev/*)$(echo /dev/*/*)$(echo /dev/*/*/*)
+	blockdevs=$(echo /dev/disk/*/*)$(lsblk -n -o name -p -l)
 
 	if [[ ${cur} == -* ]] ; then
 		COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
--- a/etc/bash_completion.d/stratis
+++ b/etc/bash_completion.d/stratis
@@ -1,23 +1,27 @@
 _stratis()
 {
-	local cur prev x2prev x3prev opts root_subcommands pool_subcommands fs_subcommands blockdev_subcommands daemon_subcommands complete_pools
+	local cur prev x2prev x3prev opts root_subcommands pool_subcommands fs_subcommands blockdev_subcommands daemon_subcommands complete_pools blockdevs first second third
 	COMPREPLY=()
 	cur="${COMP_WORDS[COMP_CWORD]}"
 	prev="${COMP_WORDS[COMP_CWORD-1]}"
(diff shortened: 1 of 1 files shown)


mulkieran:
Thanks! @agrover, @tasleson review?

etc/bash_completion.d/stratis
```diff
@@ -1,17 +1,21 @@
...
+	first="${COMP_WORDS[1]}"
+	second="${COMP_WORDS[2]}"
+	third="${COMP_WORDS[3]}"
 	opts="-h --help --version"
 	root_subcommands="--help --version daemon pool blockdev filesystem"
 	pool_subcommands="create list rename destroy"
 	fs_subcommands="create snapshot list rename destroy"
 	blockdev_subcommands="add list"
 	daemon_subcommands="redundancy version"
+	blockdevs=$(ls -l /dev | awk '{printf("/dev/%s\n",$10)}' | tr '\n' ' ')
```
agrover: Looks good to me. The only question I have (and please understand I am not very familiar with bash) is if this handles blockdev completions down into dirs under dev e.g. /dev/disk/by-id/ or not?

carzacc: That doesn't do it, I changed it and now I removed the useless awk (I found out echo actually does what I needed already) and now it also looks in subdirs and subdirs of subdirs so that it finds devices under /dev/disk/by-X/

tasleson:
A couple of thoughts (although it's been 3 years since I've worked on bash completion)...

1. In general I believe it's good practice to always refer to variable with `${name}` syntax, so `${x2prev}` instead of `$x2prev` etc.
2. For `blockdevs`, wondering if it would it be better to leverage `lsblk`, so that the user only has to pick from devices that are actual block devices?  Something like `blockdevs=$(lsblk -n -o name -p -l)` maybe?  Yes this does add an additional dependency...

agrover:
@tasleson I think stratisd is already depending on things that depend on util-linux, which provides lsblk, so using it should be fine.

I'd be ok either if this PR went ahead and used it, or we could also open a new issue and tackle it as a further enhancement in the future. I'm not sure how easy it is to parse lsblk output.

tasleson:
@agrover With what I provided I think that's it, it produces a list of block devices, so no parsing should be needed.

agrover:
I appear to be experiencing partial screen blindness today 😀

carzacc:
The only thing `$(lsblk -n -o name -p -l)` wouldn't include that `$(echo /dev/*)$(echo /dev/*/*)$(echo /dev/*/*/*)` includes that we care about are /dev/disk/by-*/ entries that @agrover mentioned in his review of the code, another way would be to have `blockdevs=$(echo /dev/disk/*/*)$(lsblk -n -o name -p -l)` so that we get all the block devices listed by lsblk and all entries in /dev/disk/by-X*/

tasleson:
@carzacc Yes, that would be a good way to handle blockdevs and the alias's.

agrover:
me too! Merging! 👍

Issue #99, opened 2017-12-18
completions for blockdevs when using pool create


agrover:
i.e. typing `stratis pool create pool1 /dev/sd<TAB>` should show all entries in /dev that start with "sd".

PR #104, opened 2018-01-10
Add a space to separate the last entry from /dev/disk/by-*/ and the first from lsblk

carzacc:
Without this change there is nothing separating the last /dev/disk/by-uuid/ entry and the first blockdev name given by lsblk so the script treats these two entries as one


agrover:
👍

carzacc:
By the way #99 should probably be closed

agrover:
Thanks again, @carzacc !

Issue #101, opened 2017-12-19
listing filesystems should show more information


agrover:
such as: created date, mount points, amount of data on the fs.

We just list the name right now, and it looks pretty spartan.

mulkieran:
@agrover We already export the devnode() on the D-Bus, should we be reporting that?

agrover:
@mulkieran the /dev/dm-X devnode? I'd rather have people use the symlink under /dev/stratis (once #682 is merged), so maybe we can show that instead..

mulkieran:
Showing the symlink is a stratisd issue first. The problem is that the symlinks are only synthesized in the source that makes them. So a refactoring is necessary to synthesize them outside of that context.

agrover:
Also, what its path in `/dev/stratis` is, for better discoverability and cut-n-pastability.

agrover:
In looking to provide the symlink, because of the implementation it's kind of hard. StratFilesystem doesn't know its own name (or uuid), and doesn't know its pool's name either.

D-Bus Filesystem API methods know their pool's name, so we could implement a call to some devlinks function from there, or we could synthesize it entirely in the cli. It really depends how confident we are that these links are not going to change with any frequency.

agrover:
fixed by #185

Issue #100, opened 2017-12-19
Things (pools, filesystems, blockdevs) should be listed in alphabetic order


agrover:
by default, at least.

mulkieran:
alphabetical by what key?

agrover:
I think by name.

mulkieran:
Resolved by #105.

PR #105, opened 2018-01-16
Add alphabetical sorting

carzacc:
This addresses #100, obviously filesystem sorting will have to be changed when #101 is addressed.

### DIFF
--- a/src/stratis_cli/_actions/_logical.py
+++ b/src/stratis_cli/_actions/_logical.py
@@ -30,6 +30,7 @@ from ._data import Pool
 from ._data import Filesystem
 from ._data import filesystems
 from ._data import pools
+from ._formatting import print_table
 
 
 class LogicalActions(object):
@@ -79,8 +80,20 @@ class LogicalActions(object):
            managed_objects,
            props={'Pool': pool_object_path}
         )
-        for _, info in matching_filesystems:
-            print(MOFilesystem(info).Name())
+
+        tables = [
+           [
+              MOFilesystem(info).Name(),
+           ] for _, info in matching_filesystems
+        ]
+
+        print_table(
+           [
+              'Name'
+           ],
+           sorted(tables, key=lambda entry: entry[0]),
+           ['<']
+        )
 
         return
 
--- a/src/stratis_cli/_actions/_physical.py
+++ b/src/stratis_cli/_actions/_physical.py
@@ -79,7 +79,7 @@ class PhysicalActions(object):
                 "Physical Size",
                 "State",
             ],
-            tables,
+            sorted(tables, key=lambda entry: entry[0]),
             ['<', '>', '>']
         )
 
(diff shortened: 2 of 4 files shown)


agrover:
LGTM 👍

src/stratis_cli/_actions/_logical.py
```diff
@@ -79,8 +79,13 @@ def list_volumes(namespace):
            managed_objects,
            props={'Pool': pool_object_path}
         )
+        fs_names = []
```
mulkieran: This would be nicer if more functional, like:

```
fs_names = \
   [name for name in map(lambda p: MOFilesystem(p[1]).Name(), matching_filesystems)]
for fs in sorted(fs_name):
    print(fs)
```

Also, I'm not sure if key should really be lowercase version of str. Might as well use the same key as their actual value.

src/stratis_cli/_actions/_physical.py
```diff
@@ -66,12 +66,13 @@ def list_pool(namespace):
...
         tables = [
             [
-                modev.Devnode(),
-                str(Range(modev.TotalPhysicalSize(), SECTOR_SIZE)),
-                state_val_to_string(modev.State()),
-            ] for modev in modevs
+                sorteddev.Devnode(),
+                str(Range(sorteddev.TotalPhysicalSize(), SECTOR_SIZE)),
+                state_val_to_string(sorteddev.State()),
+            ] for sorteddev in sorteddevs
```
mulkieran: The effect is much the same right now, but it seems better to make the table and then sort. It's easy to imagine wanting to sort on multiple fields of the table in some later version. So, revert to making tables as in original,
and then:

```
print_table(
     ...,
     [e for e in sorted(tables, key=lambda entry: entry[0]],
     ...,
)
```

src/stratis_cli/_actions/_top.py
```diff
@@ -75,12 +75,13 @@ def list_pools(_):
 
         managed_objects = ObjectManager.Methods.GetManagedObjects(proxy, {})
         mopools = (MOPool(info) for _, info in pools(managed_objects))
+        sortedpools = sorted(mopools, key=lambda po: str.lower(po.Name()))
```
mulkieran: Same idea as for ```list_pools```.

src/stratis_cli/_actions/_logical.py
mulkieran: Given that, as you point out, #101 will make this more like the others, might as well set it up for call to ```print_table``` in the same manner as the others.

mulkieran:
Travis is back to normal.

carzacc:
I have made the requested changes

mulkieran:
Looks fine to me. @agrover review, please?

carzacc:
Actually we need to change the bash completion script so that it ignores the first line of the filesystem list command

agrover:
looks good on its face, let's merge it and go from there. Thanks!

PR #116, opened 2018-03-21
Do not disable bad-continuation warnings for pylint

mulkieran:
Make a few edits so that the code will satisfy both pylint and yapf.

bad-continuation was disabled by me so that I could use my own formatting
style. Now that we are using yapf there is no point, so might as well
just conform to both. I think the code looks slightly better this way.

### DIFF
--- a/check.py
+++ b/check.py
@@ -6,17 +6,17 @@ import sys
 
 arg_map = {
     "src/stratis_cli": [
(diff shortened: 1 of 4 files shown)


agrover:
I'm gonna go out on a limb and just merge this. Thanks!

PR #117, opened 2018-03-26
Do not include rc in StratisCliRuntimeError.__str__

agrover:
This message is printed on the command-line when a call to stratisd fails
for a non-dbus-related reason. For example, when trying to create a pool
with a name that is already used by an existing pool.

The rc value is not decoded. I suppose it could be, but it seems like just
printing the message is likely to best convey what happened to the user.


agrover:
Closing. I think a better approach is to keep the rc but decode it. see https://github.com/stratis-storage/stratis-cli/issues/118

Issue #83, opened 2017-11-02
We can generate more user friendly messages from exceptions


mulkieran:
The only information we have when the exception finally propagates to the top is the exception chain.

We should set up a framework so that other people can contribute code that inspects the error chain and comes up w/ friendlier error messages than what we show now.

agrover:
This seems to lead into supporting localized strings and error messages in the future.

mulkieran:
A first step would be two have a ```rich-error``` flag which would just traverse the exception chain, printing out error messages in turn, as:

```
Error 1: Failed to make stratisd D-Bus connection.
Error 2: The name org.storage.stratis1 was not provided by any .service files
```

or something.

mulkieran:
This gets to seem more urgent, because some of our errors do not have a string representation at all. This is the origin of
"Execution failed:" with no following message stratis response: https://bugzilla.redhat.com/show_bug.cgi?id=1563561.

Issue #70, opened 2017-09-29
Our goal is to catch all internal dbus-python errors, raising our own with the internal error as cause


mulkieran:
We can work on doing this gradually.

mulkieran:
Handling DPClientRuntimeErrors...
These can be reraised as a result of
  *. DBusExceptions being raised on an actual call, a Get, a Set, or a method call
  *. IntoDPError being raised if arguments are not transformed properly. Most likely this would be a result of caller arguments being badly constructed.
Or raised because of:
  *. A failure by the caller to use the correct keys for method arguments

Should dbus-python-client-gen do more differentiating of errors or should stratis-cli just work w/ what it gives?

I think subtyping in dbus-python-client-gen might be the correct thing to do. Maybe:

* DPClientInvocationError
  * DPClientCallError
  * DPClientSetPropertyError
  * DPClientGetPropertyError

*DPClientInvalidArgError ---
  * DPClientInvalidKeyWordError --- for caller just specified wrong keywords
  * DPClientMarshallingError --- for IntoDPError

mulkieran:
AFAIK, the only dbus-python error that could be propagated directly are in _connection.py. Those are starting the bus and getting an object.

mulkieran:
See: https://github.com/stratis-storage/dbus-python-client-gen/pulls.

mulkieran:
All the stratisd daemon specific errors will be raised due to the ```get_object``` action, not due to the bus getting, which is only D-Bus or maybe system bus specific.

mulkieran:
Is it reasonable to wrap every action method? The additional information that this would give is the namespace passed to the parser.

All the wrapper would do is add a try/except block which caught the resulting exceptions, which would be one of:

```dbus.exceptions.DBusException, DPClientRuntimeError, StratisCliError``` and reraise with ActionError, subclass of RuntimeError.

ActionError contains two fields, function name, and namespace, or possibly just namespace. Something like:

```
def wrapper(func):
  def the_func(namespace):
    try:
      func(namespace)
   except (dbus.exceptions.DBusException, DPClientRuntimeError,
                StratisCliError) as err:
    raise StratisCliActionError(namespace) from err

  return the_func
```

[...]

PR #124, opened 2018-04-04
Try to make CLI errors a bit more interpretable

mulkieran:
Resolves: #70
Resolves: #83

### DIFF
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -1,0 +1,67 @@
+# Copyright 2016 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Facilities for managing and reporting errors.
+"""
+import os
+import sys
+
+import dbus
+
+
+def get_errors(exc):
+    """
+    Generates a sequence of exceptions starting with exc and following the chain
+    of causes.
+    """
+    while True:
+        yield exc
+        exc = getattr(exc, "__cause__") or getattr(exc, "__context__")
+        if exc is None:
+            return
+
+
+def get_error_msgs(errors):
+    """
+    Generates messages from a sequence of errors. Omits empty messages
+    """
+    for error in errors:
+        if isinstance(error, dbus.exceptions.DBusException):
+            error_str = error.get_dbus_message()
+        else:
+            error_str = str(error)
+        if error_str is not None and error_str != "":
+            yield error_str
+    return
+
+
+def handle_error(err):
+    """
+    Do the right thing with the given error, which may be the head of an error
+    chain.
+
+    :param Exception err: an exception
+    """
+
+    error_msgs = [msg for msg in get_error_msgs(get_errors(err))]
+    if error_msgs == []:
+        # It is unlikely that, within the whole chain of errors, there
+        # will be no message that is not an empty string. If there is
+        # there is some program error, so just raise the exception.
+        raise err
+
+    error_msg = ("%s    which in turn caused:%s" %
+                 (os.linesep, os.linesep)).join(reversed(error_msgs))
+
+    sys.exit("Execution failure caused by:%s%s" % (os.linesep, error_msg))
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -54,7 +54,13 @@ def handle_error(err):
     :param Exception err: an exception
     """
 
-    error_msgs = [msg for msg in get_error_msgs(get_errors(err))]
+    errors = [error for error in get_errors(err)]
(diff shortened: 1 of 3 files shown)


mulkieran:
Simple example:

```
[mulhern@localhost stratis-cli]$ PYTHONPATH=./src ./bin/stratis pool list
Execution failed; causes in order of specificity:
Error 0: The name org.storage.stratis1 was not provided by any .service files
```

mulkieran:
Old style:

```
[mulhern@localhost stratis-cli]$ PYTHONPATH=./src ./bin/stratis pool create test1 bd
Execution failed: 
```
it caught a DPClientRuntimeError which has str representation "" and just printed that.

New way:

```
[mulhern@localhost stratis-cli]$ PYTHONPATH=./src ./bin/stratis pool create test1 bd
Execution failed; causes in order of specificity:
Error 0: Invalid argument 0
```

Traversed the cause chain, printing out string representation of errors that have one.

The error is artificially generated. I changed the data so that stratis-cli exported a boolean for the value of the pool name, so the error string is correct.

agrover:
I like it. Slight bikeshed -- replace "causes in order of specificity" with "caused by"?

mulkieran:
Ok. String is now "Execution failure caused by:"

mulkieran:
Here is another induced error, two item chain:

```
[mulhern@localhost stratis-cli]$ PYTHONPATH=./src ./bin/stratis pool create test1 bd
Execution failure caused by:
Error 0: value 'test1' for parameter expr is unacceptable: could not be transformed
Error 1: invalid literal for int() with base 10: 'test1'
```

The induced error was that the name of the pool should be transformable to an int. So we get the actual exception from ```int(test1)``` as Error 1, and then the exception from the ```into-dbus-python``` library method as Error 0.

mulkieran:
Ok. I think this is ready to go.

agrover:
Just to confirm my understanding here, if we `raise` because --propagate was given, if we have three chained exceptions we'll see the (messy but useful to devs) three separate stack traces, right?

On another topic, I found myself getting hung up on the "Execution failed but no printable cause" string. Is this really needed? When do we see str(error) fail? If never, do we need this? If sometimes, is using `error.__name__` as a fallback enough to guarantee we always have something, and thus can avoid this scenario?

Finally, I'm worried looking at the example given that it's not clear that Error 1 caused Error 0, and not vice versa, if users misinterpret the ordering. Users may also miss the causal relationship between the two, and think that the command generated two separate errors instead of one linked error. Given this, I would recommend reversing the order so highest-level exception is given last (would users tend to look at the last thing first?), and inserted "caused" (or "caused by" if you decide to stick with the existing ordering) between lines to help the user understand the ordering better from context.

Finally finally (hey, I think this stuff is really important), It might be nice to number errors starting with 1 instead of 0.

mulkieran:
> Just to confirm my understanding here, if we raise because --propagate was given, if we have three chained exceptions we'll see the (messy but useful to devs) three separate stack traces, right?

Correct.

mulkieran:
> On another topic, I found myself getting hung up on the "Execution failed but no printable cause" string. Is this really needed? When do we see str(error) fail? If never, do we need this? If sometimes, is using error.__name__ as a fallback enough to guarantee we always have something, and thus can avoid this scenario?

It is unlikely that we'll ever have an empty string for the error with these changes, but who knows with Python? How about just propagating the exception when that happens? That should make it more likely that it will be reported as a bug, and if reported, that we can fix it.

agrover:
Sounds good to me.

mulkieran:
Here is an unnumbered approach:

```
[mulhern@localhost stratis-cli]$ PYTHONPATH=./src ./bin/stratis pool create test1 junk
Execution failure caused by:
invalid literal for int() with base 10: 'test1'
    which in turn caused
value 'test1' for parameter expr is unacceptable: could not be transformed
```

mulkieran:
With enhanced dbus-python-client-gen messages:

```
[mulhern@localhost stratis-cli]$ PYTHONPATH=./src ./bin/stratis  pool create 1 junk
Execution failure caused by:
Invalid argument 0
    which in turn caused
Error while invoking method "CreatePool" belonging to interface "org.storage.stratis1.Manager" with arguments (1, dbus.Struct((dbus.Boolean(True), dbus.UInt16(0)), signature=dbus.Signature('bq')), 0, dbus.Array([dbus.String('junk')], signature=dbus.Signature('s')))
```

mulkieran:
Another dbus-python-client-gen "enhanced" version:

```
[mulhern@localhost stratis-cli]$ PYTHONPATH=./src ./bin/stratis pool create test junk
Execution failure caused by:
invalid literal for int() with base 10: 'test'
    which in turn caused
value 'test' for parameter expr is unacceptable: could not be transformed
    which in turn caused
Failed to format arguments (test, (True, 0), False, ['junk'])
```

agrover:
seems like an improvement to me!

mulkieran:
rebased.

mulkieran:
Commits squashed, code unchanged.

mulkieran:
Might as well try the ActionError idea in #70.

mulkieran:
With our one interpretation string:

```
[mulhern@localhost stratis-cli]$ PYTHONPATH=./src ./bin/stratis daemon version
Execution failure caused by:
Could not get owner of name 'org.storage.stratis1': no such name
    which in turn caused
The name org.storage.stratis1 was not provided by any .service files

Most likely the Stratis daemon, stratisd, is not running.
```

agrover:
looks very good. nit: Please put a ':' after "which in turn caused"? Thanks! :+1:

mulkieran:
Ok. ':' added, moved to review final.

PR #125, opened 2018-04-13
Catch DPClientGenerationError and raise StratisCliError

mulkieran:
This is a fairly trivial change, since the exception is still raised, just
a more appropriate StratisCli error, instead of a DPClient error.

This is probably fine, because we expect to ensure well-formed introspection
data. It's also not a behavior that can be changed w/ a flag, because it
occurs during module load, that is, before any flags are read.


mulkieran:
This is virtually trivial.

agrover:
:+1:

mulkieran:
Slightly reworded FIXME comment for greater precision...

PR #127, opened 2018-04-20
Refine and improve error heirarchy just a bit

mulkieran:
Related: #7


mulkieran:
Ok. Since no objection, putting it up for final review.

PR #128, opened 2018-04-23
Correct explanation of --propagate flag in man page

mulkieran:


agrover:
I think this means to say that when debugging the daemon it's useful to use --propagate, but it seems to say that everyone including non-developer users should use it always, which I don't think is true.

mulkieran:
The Recommendation part is under sub-heading ```REPORTING BUGS AND DEVELOPMENT``` to restrict the "run with --propagate" advice to that context. I guess it could be restricted further, somehow.

mulkieran:
Ok. I'm putting just the non-controversial bit up for review at this time.

agrover:
> Allow any exceptions raised during execution to propagate.

I don't think this is quite accessible enough to non-programmers. The words that are tripping me up are "exceptions" -- can we say "errors"? -- and "propagate". I'm assuming the user is reading this because they don't know what propagate means in this context, so best to use other word choices in the explanation.

mulkieran:
I think saying what the effect of the flag actually is is necessary. I added a bit to make it sound like a "developer" kind of thing, not for typical users.

agrover:
I'm saying I, as a developer, still don't know what "Allow any exceptions raised during execution to propagate" means. Looking at the code it appears to just mean printing the exception/stacktrace instead of just printing the error. If so why is "--propagate" a better description than something like "--verbose-errors"?

mulkieran:
It means allow the exception to propagate; leave it to the interpreter to decide what to do when it catches the exception.

mulkieran:
@agrover review?

mulkieran:
Are you telling me that it is not the intepreter that prints the exception, but the shell?

mulkieran:
New commit message, line, and ordering.

@agrover review, please?

agrover:
love it. merging.

PR #130, opened 2018-04-25
Clean up unit tests a bit

mulkieran:
Remove class heirarchy in __init__.py. There's no reason for it anymore,
since the C version is now obsolete.

Have Service.setUp() call time.sleep(). This is an improvement over having
it called after every call to Service.setup().

All these things have been done in stratisd's dbus-client tests already.


agrover:
:+1:

PR #139, opened 2018-05-07
Allow lint task to fail

mulkieran:
It requires dbus-python which has increased the version of dbus-1 which
it depends on from 1.6 to 1.8.  Travis's version of Ubuntu does not
currently supply that version.


mulkieran:
This is all temporary, of course.

PR #143, opened 2018-05-09
Pin dbus-python to 1.2.4

mulkieran:
This allows running the lint test on Travis while Travis's version of
Ubuntu does not support libdbus v.1.8. This is the most recent available
version of dbus-python which is not the current, problematical one that
requires libdbus v.1.8. When Travis has upgraded sufficiently, this pin
can be dropped.

No longer allow lint task to be a failure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,6 @@ branches:
     only: master
 
 matrix:
-    allow_failures:
-        - env: TASK=lint
     include:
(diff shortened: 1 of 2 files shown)


mulkieran:
If everyone is comfortable with this solution to our lint task failure, I will put it in elsewhere as necessary.

agrover:
:+1:

Issue #145, opened 2018-05-11
We should be very paranoid and catch AttrbuteErrors as well


mulkieran:
They can always arise on a mis-spelled method, and would not be noticed until run time. We might be able to distinguish between errors in our code and errors in other's code, but that is about it.

We should report it in our richer error reporting.

mulkieran:
AttributeErrors themselves do not have a lot of specfic data, otherwise we could check the type on which the error was reported.

PR #147, opened 2018-05-21
Also catch and process an AttributeError

mulkieran:
Resolves: #145

### DIFF
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -59,6 +59,12 @@ def interpret_errors(errors):
     # pylint: disable=fixme
     # TODO: This method is extremely rudimentary. It should not be extended
     # using exactly the structure it has now.
+    if isinstance(errors[1], AttributeError):
+        import traceback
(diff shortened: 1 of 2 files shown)


agrover:
> "There seems to be a typo in the source at line %d in file "
                   "%s. The offending line is \"%s\".")

Can this be terser? "Code error at line %d in file %s: \"%s\"."

mulkieran:
These are explanations, so I don't really think we should go for terseness. But typo is too reductionist, so often the problem is that the type is wrong. I've changed the explanation to make it less likely to be wrong.

agrover:
I'd like to omit "most likely" because don't we know there definitely is an error?

I'd like to omit "offending" because it's somewhat folksy.

mulkieran:
We know there is an error, but we are only making an informed guess about the cause of the error. That is why "Most likely".

mulkieran:
"offending" isn't folksy, but it is not necessary either. I can leave it out.

agrover:
ok, I'm done bikeshedding :smiley:

PR #149, opened 2018-05-30
Catch all exception when trying to interpret an error

mulkieran:
Try to avoid causing an exception raised while trying to interpret an error
to cause all error reporting to fail.

If trying to interpret the error causes an exception, then we have no
interpretation of the error, clearly, so returning None in that case is
a reasonable response.


mulkieran:
rebased.

agrover:
:+1:

Issue #131, opened 2018-04-26
RFE: Provide zsh completions


myllynen:
It would be nice if zsh completions could be provided for stratis-cli.

In the past I've authored libvirt and openstack completions for zsh which both support a huge amount of subcommands and options but the completions even though supporting all of them are relatively compact so perhaps a similar type approach could be considered with stratis-cli as well (this might require some changes to the client output), I'm providing these two zsh completions here for reference:

https://sourceforge.net/p/zsh/code/ci/master/tree/Completion/Unix/Command/_libvirt
https://sourceforge.net/p/zsh/code/ci/master/tree/Completion/Unix/Command/_openstack

If you are interested to make users' life even more easier and push things further, much more intelligent completions with zsh are also possible but that naturally would require more effort, however e.g. the libvirt completions above already demonstrate how to provide smarter completions for a set of subcommands.

Thanks.

mulkieran:
Is #106 relevant here?

myllynen:
Yes, I think so. If you're planning changes on this area perhaps checking virsh and openstack clients for inspiration might be a good idea (in essence, both provide subcommand and options in a standard format that's pretty straightforward to parse in completion functions). Thanks.

PR #152, opened 2018-06-14
Add completion function for Zsh

okapia:
To complement the completion for bash, here is a function for zsh. I've placed it in the etc directory but I don't know whether you perhaps want it elsewhere. It should be installed in `/usr/share/zsh/site-functions` so that zsh will pick it up when compinit is run.

When testing stratis, stratis pool list returned "TypeError: the_func() got multiple values for argument 'props'". So for the pool completion, I've had to believe the bash completion.

Resolves: #131.

### DIFF
--- a/shell-completion/zsh/_stratis
+++ b/shell-completion/zsh/_stratis
@@ -1,0 +1,89 @@
+#compdef stratis
+
+# Completion function for zsh
+
+(( $+functions[_stratis_pools] )) ||
+_stratis_pools() {
+  local -a expl
+  _description pools expl pool
+  compadd "$@" "$expl[@]" - \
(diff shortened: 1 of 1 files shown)


okapia:
That sounds like a good approach to me.

mulkieran:
@okapia: To get rid of those errors, update dbus-client-gen to v0.3.

agrover:
@okapia cool. Would you be so kind as to revise this PR to use the new location?

okapia:
Revised.

(A rare opportunity to re-familiarize myself with git filter-branch).

mulkieran:
I'm opting out of review, this is not my area.

agrover:
Merged. Thanks much!

PR #153, opened 2018-06-15
Move bash shell completion under shell-completion/bash

agrover:
Now that we have two shell completion scripts, put them in one place.


mulkieran:
I don't feel I can comment on this one.

agrover:
OK, well let's proceed.

PR #160, opened 2018-07-17
Increase timeout over default

tasleson:
The stratisd daemon will at times exceed the default timeout of
~25 seconds if we allow it to create more than 1 FS at a time.
At this time we are advocating that the timeout be set to
something much longer to reduce the probability of exceeding
the timeout.


agrover:
I'm sure the substance of this is fine, but there seems to be a number of other (automatic?) changes. I'm sure are they are also uncontroversial, but please put them in a separate PR? Thanks.

agrover:
Actually I think the just-merged #161 fixed the same lints, so if this just gets rebased we should be good to go.

tasleson:
@agrover Re-based

tasleson:
Constant moved to `_actions/_constants.py`

Issue #164, opened 2018-07-31
Cryptic cli error when you don't have adequate privileges


tasleson:
```$ stratis pool create fubar /dev/sde
Execution failure caused by:
Rejected send message, 2 matched rules; type="method_call", sender=":1.55" (uid=17209 pid=9356 comm="/usr/bin/python3 -s /usr/bin/stratis pool create f" label="unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023") interface="org.storage.stratis1.Manager" member="CreatePool" error name="(unset)" requested_reply="0" destination=":1.53" (uid=0 pid=9345 comm="target/debug/stratisd --debug " label="unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023")
```

mulkieran:
If possible, please run again w/ --propagate flag set.

tasleson:
As requested
```
[tasleson@localhost stratis-cli]$ bin/stratis --propagate pool create fubar /dev/sde
Traceback (most recent call last):
  File "/home/tasleson/projects/stratis/dbus-python-client-gen/src/dbus_python_client_gen/_invokers.py", line 319, in dbus_func
    return dbus_method(*xformed_args, timeout=timeout)
  File "/usr/lib64/python3.6/site-packages/dbus/proxies.py", line 145, in __call__
    **keywords)
  File "/usr/lib64/python3.6/site-packages/dbus/connection.py", line 651, in call_blocking
    message, timeout)
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.AccessDenied: Rejected send message, 2 matched rules; type="method_call", sender=":1.127" (uid=17209 pid=9793 comm="python3 bin/stratis --propagate pool create fubar " label="unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023") interface="org.storage.stratis1.Manager" member="CreatePool" error name="(unset)" requested_reply="0" destination=":1.53" (uid=0 pid=9345 comm="target/debug/stratisd --debug " label="unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023")

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/tasleson/projects/stratis/stratis-cli/src/stratis_cli/_main.py", line 42, in the_func
    result.func(result)
  File "/home/tasleson/projects/stratis/stratis-cli/src/stratis_cli/_actions/_top.py", line 57, in create_pool
[...]

PR #165, opened 2018-08-01
Extend and improve interpretation of D-Bus errors

mulkieran:
Adds an interpretation for ```AccessDenied``` error. Triggers "Most likely the Stratis daemon, stratisd, is not running" message on ```NameHasNoOwner``` exception, rather than ```ServiceUnknown``` exception which is higher level and less specific.

Resolves: #164 .

### DIFF
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -74,24 +74,26 @@ def interpret_errors(errors):
     # TODO: This method is extremely rudimentary. It should not be extended
     # using exactly the structure it has now.
     try:
-        if isinstance(errors[1], AttributeError):
+        # Inspect top-most error after StratisCliActionError
+        error = errors[1]
+        if isinstance(error, AttributeError):
             import traceback
-            frame = traceback.extract_tb(errors[1].__traceback__)[-1]
+            frame = traceback.extract_tb(error.__traceback__)[-1]
             fmt_str = (
                 "Most likely there is an error in the source at line %d "
                 "in file %s. The text of the line is \"%s\".")
             return fmt_str % (frame.lineno, frame.filename, frame.line)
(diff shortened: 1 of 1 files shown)


mulkieran:
```
[mulhern@localhost stratis-cli]$ PYTHONPATH=./src ./bin/stratis pool create fubar /dev/sde
Execution failure caused by:
Rejected send message, 2 matched rules; type="method_call", sender=":1.855" (uid=1000 pid=21926 comm="python3 ./bin/stratis pool create fubar /dev/sde " label="unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023") interface="org.storage.stratis1.Manager" member="CreatePool" error name="(unset)" requested_reply="0" destination=":1.852" (uid=0 pid=21858 comm="./target/x86_64-unknown-linux-gnu/debug/stratisd -" label="unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023")
    which in turn caused:
Error while invoking method "CreatePool" belonging to interface "org.storage.stratis1.Manager" with arguments (fubar, dbus.Struct((dbus.Boolean(True), dbus.UInt16(0)), signature=dbus.Signature('bq')), 0, dbus.Array([dbus.String('/dev/sde')], signature=dbus.Signature('s')))

Most likely you have insufficient permissions for the action you specified.
```

tasleson:
This is an improvement , but if our goal is simplicity should we output the information before the message `Most likely you have insufficient permissions for the action you specified`?  Perhaps that could be outputted if a user specifies an additional parameter for more verbose errors?

mulkieran:
We have an infrastructure and format that we have agreed on. There certainly are ways that we can modify that format and infrastructure, but that discussion is really orthogonal to this problem. Please feel free to start a separate issue about altering the format with which we present errors to the user.

PR #175, opened 2018-08-14
Change 'stratis pool' and 'stratis blockdev' behavior

agrover:
Change from showing subcommand help to "list" behavior. Help is still available for subcommands by giving `--help`. This should make the default behavior a little friendlier.

PR #176, opened 2018-08-14
Test listing blockdevs when no pool is specified

mulkieran:
This also does an explicit comparison when checking whether there has been a pool name specified.


mulkieran:
Blocks #175.

src/stratis_cli/_actions/_physical.py
```diff
@@ -73,7 +73,7 @@ def list_pool(namespace):
 
         # If user specified a pool name we wil constrain output to that one
         # pool, else we will output all blockdevs.
-        if namespace.pool_name:
+        if namespace.pool_name is not None:
```
mulkieran: This collides with #175 change.

PR #180, opened 2018-08-15
Enhance output of 'stratis fs list'

agrover:
* Based on example of previous work on blockdevs, do not require pool name to be given.
* Add output of newly added fields: Created date, used space, and mount points.
* Also print Devnode, which is now `/dev/stratis/<poolname>/<fsname>`.

Requires additional dependency on python-dateutil.

### DIFF
--- a/src/stratis_cli/_actions/_logical.py
+++ b/src/stratis_cli/_actions/_logical.py
@@ -87,10 +87,9 @@ class LogicalActions():
             mofilesystem.Devnode(),
         ] for mofilesystem in mofilesystems]
 
-        print_table(
-            ['Pool Name', 'Name', 'Used', 'Created', 'Device'],
-            sorted(tables,
-                   key=lambda entry: entry[0]), ['<', '<', '<', '<', '<'])
+        print_table(['Pool Name', 'Name', 'Used', 'Created', 'Device'],
+                    sorted(tables, key=lambda entry: entry[0]),
+                    ['<', '<', '<', '<', '<'])
 
     @staticmethod
     def destroy_volumes(namespace):
--- a/src/stratis_cli/_actions/_logical.py
+++ b/src/stratis_cli/_actions/_logical.py
@@ -17,8 +17,8 @@ Miscellaneous logical actions.
 
 from __future__ import print_function
 
-from justbytes import Range
 from dateutil import parser as date_parser
+from justbytes import Range
 
 from .._errors import StratisCliEngineError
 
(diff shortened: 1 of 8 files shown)


agrover:
Example output of `stratis fs list`:
```
Pool Name  Name  Used        Created            Device                  Mount Points  
test1      fs1   > 1.05 GiB  Aug 10 2018 12:41  /dev/stratis/test1/fs1  /root/mnt, /root/mnt2
```

mulkieran:
> Upon reflection, I think maybe we don't want to list mount points. One, because it's potentially a list, which is hard to show in fs list, and also because this information is already available with the mount command that users are already likely familiar with. Thoughts?

Less is more, AFAIC.

agrover:
> Also, please run up some tests as well. test/integration/logical/test_list.py is probably the correct place.

Any ideas for good things to test here?

agrover:
OK, added stuff to test, as much as we can.

src/stratis_cli/_actions/_logical.py
```diff
@@ -66,20 +70,26 @@ def list_volumes(namespace):
...
+            MOFilesystem(info) for _, info in filesystems(props=properties)
+            .search(managed_objects)
+        ]
 
-        print_table(['Name'], sorted(tables, key=lambda entry: entry[0]),
-                    ['<'])
+        tables = [[
+            path_to_name[mofilesystem.Pool()],
+            mofilesystem.Name(),
+            str(Range(mofilesystem.Used(), 1)),
```
mulkieran: ```str(Range(mofilesystem.Used()))``` is better. I think the "1" can not do any harm, but it should be entirely unnecessary.

src/stratis_cli/_parser/_parser.py
```diff
@@ -18,6 +18,7 @@
 import argparse
 
 from .._actions import PhysicalActions
+from .._actions import LogicalActions
```
mulkieran: Please reverse order, "L" before "P".

tests/integration/logical/test_list.py
```diff
@@ -123,3 +123,53 @@ def testList(self):
...
+                        ] + self._VOLUMES[2]
+        RUNNER(command_line)
+
+    def tearDown(self):
+        """
+        Stop the stratisd simulator and daemon.
+        """
+        self._service.tearDown()
+
+    def testList(self):
```
mulkieran: If you are only going to use ```_MENU``` in one place you might as well get rid of it entirely and just write it out explicitly in first line.

agrover:
cool, thanks. will do.

PR #181, opened 2018-08-16
Augment an error diagnosis

mulkieran:
As we have seen, we do not have, at this time, any way to disambiguate
between stratisd not running at all and stratisd running but no D-Bus
connection established.

It would be possible to do better, by actually checking.


agrover:
Can we be more succinct? How about "Cannot connect to the Stratis D-Bus service". This translates the arcane NameHasNoOwner exception, but attempts to diagnose beyond that may be more than what's required.

mulkieran:
Looks like a bogus Travis error.

Issue #172, opened 2018-08-09
Non-intuitive  add-data add-cache should be options on a pool not a blockdev


tasleson:
When I think about adding a block device to an existing pool my first thought is that it should be an option on the pool, not the blockdev.  This seems very non-intuitive to me.

Today
``` bash
# stratis blockdev add-data some_pool_name /dev/sde
```

Expected
``` bash
# stratis pool add-data some_pool_name /dev/sde
```

This would also be consistent with the actual dbus API too, it has these methods are on the pool object.

PR #193, opened 2018-09-12
Move add-data & add-cache operations to pool

tasleson:
Fixes: https://github.com/stratis-storage/stratis-cli/issues/172


tasleson:
I really need to enable the ability to run all these tests locally

PR #195, opened 2018-09-20
Remove 'force' option

tasleson:
We have removed the 'force' parameter in the API call.  The API
caller is now responsible for clearing any signatures before calling
the Stratis dbus API which consumes a block device.

Fixes https://github.com/stratis-storage/stratisd/issues/613

### DIFF
--- a/docs/stratis.txt
+++ b/docs/stratis.txt
@@ -31,7 +31,7 @@ GLOBAL OPTIONS
 
 COMMANDS
 --------
-pool create [--force] <pool_name> <blockdev> [<blockdev>..]::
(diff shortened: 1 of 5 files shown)


mulkieran:
fmt errors need to be fixed before my ack will count. Otherwise, this is fine w/ me.

tasleson:
@mulkieran @agrover Corrected formatting

agrover:
Merged stratisd PR, let's try restarting the Jenkins and seeing if it passes.

agrover:
that was fast.

Issue #204, opened 2018-10-08
stratis-cli can verify that it is running against the correct version of stratisd


mulkieran:
Packaging should ensure that stratis-cli is always installed against the correct version of stratisd, but a dynamic check wouldn't hurt.

mulkieran:
You only want to check this when actually communicating w/ stratisd. So, ```stratis --version``` does not get on the D-Bus and it shouldn't cause this check to occur.

mulkieran:
Also, it is necessary to know what version of stratisd this version of stratis requires. This is not, I think, anywhere in stratis-cli source currently.

mulkieran:
There are only two ways to go about this that I can see:

1. Put in an extra method that checks whether the version of stratisd is compatible. This would work and is simple, but likely most users wouldn't use it.

2. Before every stratis-cli action that actually communicates with stratisd check the version and make sure it's compatible. This has the advantage that the user doesn't have to do anything, but is a bit more complex than (1). Also, it adds a single, simple D-Bus property check to every such method. Probably lightweight compared to the method's real task, but still an addition.

I prefer the second approach, but it is more work. Also, it requires somebody to keep the stratisd version information up-to-date in the stratis-cli code base.

mulkieran:
Resolved by #219.

PR #207, opened 2018-11-08
Add some facilities about checking stratisd version

mulkieran:
Resolves: #204.

### DIFF
--- a/src/stratis_cli/_actions/_data.py
+++ b/src/stratis_cli/_actions/_data.py
@@ -29,6 +29,9 @@ from .._errors import StratisCliValueError
 
 from ._constants import DBUS_TIMEOUT_SECONDS
 
+MAXIMUM_STRATISD_VERSION = (1, 0, 1)
+REQUIRED_STRATISD_VERSION = (1, 0, 1)
+
 SPECS = {
     "org.freedesktop.DBus.ObjectManager":
     """
--- a/src/stratis_cli/_actions/_data.py
+++ b/src/stratis_cli/_actions/_data.py
@@ -32,6 +32,9 @@ from ._constants import DBUS_TIMEOUT_SECONDS
 MAXIMUM_STRATISD_VERSION = (1, 0, 1)
 MINIMUM_STRATISD_VERSION = (1, 0, 0)
 
+# a sanity check on the relation between the minimum and maximum versions
+assert MINIMUM_STRATISD_VERSION <= MAXIMUM_STRATISD_VERSION
+
 SPECS = {
     "org.freedesktop.DBus.ObjectManager":
     """
--- a/src/stratis_cli/_actions/_data.py
+++ b/src/stratis_cli/_actions/_data.py
@@ -30,7 +30,7 @@ from .._errors import StratisCliValueError
 from ._constants import DBUS_TIMEOUT_SECONDS
 
 MAXIMUM_STRATISD_VERSION = (1, 0, 1)
-REQUIRED_STRATISD_VERSION = (1, 0, 1)
+MINIMUM_STRATISD_VERSION = (1, 0, 0)
 
 SPECS = {
     "org.freedesktop.DBus.ObjectManager":
--- a/src/stratis_cli/_actions/_data.py
+++ b/src/stratis_cli/_actions/_data.py
@@ -29,7 +29,7 @@ from .._errors import StratisCliValueError
 
 from ._constants import DBUS_TIMEOUT_SECONDS
 
-MAXIMUM_STRATISD_VERSION = (1, 0, 1)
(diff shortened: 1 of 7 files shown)


mulkieran:
Possible wrapper:

```
def wrapper(orig_func):

    def the_func(namespace);
        verify_stratisd_version()
        orig_func(namespace)
    return the_func
```

That seems too good to be true.

mulkieran:
Example with ```destroy_pool```:

```
[mulhern@localhost stratis-cli]$ PYTHONPATH=./src ./bin/stratis pool destroy /dev/junk
Execution failure caused by:
stratisd version (0, 5, 2) does not meet stratis version requirement (1, 0, 1)

The version of stratisd you are using (0, 5, 2) is not compatible with this version of stratis. stratis can execute only the subset of its commands that do not require stratisd.

```

agrover:
What commands can stratis run if it's not version-compatible with stratisd? Just its own version and stratisd version commands, is that it?

mulkieran:
> What commands can stratis run if it's not version-compatible with stratisd? Just its own version and stratisd version commands, is that it? 

Those and all help commands.

src/stratis_cli/_actions/_data.py
```diff
@@ -29,6 +29,9 @@
 
 from ._constants import DBUS_TIMEOUT_SECONDS
 
+MAXIMUM_STRATISD_VERSION = (1, 0, 1)
+REQUIRED_STRATISD_VERSION = (1, 0, 1)
```
agrover: suggest `MINIMUM_STRATISD_VERSION`? And, why should this not be 1.0.0?

src/stratis_cli/_error_reporting.py
```diff
@@ -92,6 +94,13 @@ def interpret_errors(errors):
         if isinstance(error, DbusClientMissingPropertyError):
             return _DBUS_INTERFACE_MSG
 
+        if isinstance(error, StratisCliStratisdVersionError):
+            fmt_str = (
+                "The version of stratisd you are using %s is not compatible "
+                "with this version of stratis. stratis can execute only the "
+                "subset of its commands that do not require stratisd.")
```
agrover: how about something like "stratis can only run help and version-related commands" for the second sentence there?

agrover: also, maybe include min/max versions stratis-cli can handle in the output?

mulkieran: > how about something like "stratis can only run help and version-related commands" for the second sentence there?

We don't want to be too specific here, or the text will rapidly get out of date. For example, I think it's possible to also query redundancy, and in future maybe more. But this message can certainly be more informative.

src/stratis_cli/_actions/_data.py
mulkieran: > suggest `MINIMUM_STRATISD_VERSION`? And, why should this not be 1.0.0?

Yes. ```MINIMUM``` is better. About the second question, I was just guessing wildly. If you think it's 1.0.0, that's what you can have. Eventually this will be the biggest burden brought on by the code in this PR, keeping track of the minimum and maximum version. Are we all up for it?

tasleson: If we adhere to semantic versioning we should only need to worry about a minimum version

mulkieran: > If we adhere to semantic versioning we should only need to worry about a minimum version

We have explicitly not stabilized our D-Bus API yet, so I think maximum is still worth keeping track of.

mulkieran:
What a report of a version error looks like now:

```
[root@localhost stratis-cli]# PYTHONPATH=./src ./bin/stratis pool destroy junk
Execution failure caused by:
stratisd version (1, 0, 1) does not meet stratis version requirements; the version must be at least (1, 0, 0) and no more than (1, 0, 1)

stratisd version (1, 0, 1) does not meet stratis version requirements; the version must be at least (1, 0, 0) and no more than (1, 0, 1). stratis can execute only the subset of its commands that do not require stratisd.
```

mulkieran:
Travis errors are caused by new version of pylint, should be fixed in #213.

mulkieran:
Ok! Now back in review prelim. If all good I will go ahead and decorate a lot of methods.

mulkieran:
rebased to get rid of pylint errors.

mulkieran:
Ok. I will go ahead and add the rest.

mulkieran:
Jenkins tests failed because stratisd just got a new version: https://stratis-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/stratis-cli/186/console. Perfect timing for testing.

mulkieran:
Superceded by #219.

PR #209, opened 2018-11-09
Output UUID in fs list

### LINKED ISSUE #189: Output FS UUID when listing filesystems
A user can do `blkid -p /stratis/<pool name>/<fs name>` to get the FS UUID, but it might be nice to output this when they list filesystems.

carzacc:
Fixes #189

### DIFF
--- a/src/stratis_cli/_actions/_logical.py
+++ b/src/stratis_cli/_actions/_logical.py
@@ -15,8 +15,6 @@
 Miscellaneous logical actions.
 """
 
-import subprocess
-
 from justbytes import Range
 from dateutil import parser as date_parser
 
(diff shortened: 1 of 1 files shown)


agrover:
Hi thanks for the PR. Stratis fs uuid property is identical to XFS's uuid, as reported by blkid. I think we can avoid calling blkid by reading uuid property in a similar manner to the other fs properties. Can we try that approach?

carzacc:
> Hi thanks for the PR. Stratis fs uuid property is identical to XFS's uuid, as reported by blkid. I think we can avoid calling blkid by reading uuid property in a similar manner to the other fs properties. Can we try that approach?

I changed, is it fine now?

mulkieran:
@trgill Could you add the magic words to test with Jenkins?

PR #211, opened 2018-11-13
Add instructions to the man page on how to use wipefs to make a blockdev usable with Stratis

### LINKED ISSUE #208: Manpage: suggest how to use previously in-use blockdevs
Since we decided not to re-add --force (#199) we should instead mention `wipefs -a` as the preferred way to make a blockdev "unused" and suitable for Stratis use. This should be mentioned in the man page, and maybe also in a help text message when stratisd refuses to create/add a blockdev with an INUSE error.

carzacc:
Fixes #208 along with https://github.com/stratis-storage/stratisd/pull/1321


agrover:
Hey thanks @carzacc! Sneaking it in for the 1.0.2 release.

PR #212, opened 2018-11-16
Bug fix: Use the value obtained with hasattr in match

mulkieran:
This worked, because the key for the pool name in the namespace was
always "pool_name", but the whole point of having the parameter was
to avoid the dependency on a global naming scheme.


mulkieran:
This is a very limited in scope bug fix, it should be easy to ack.

PR #216, opened 2018-12-11
Update bash shell completion

carzacc:
There have been a few changes to the CLI since I've written that bash completion code, so this PR aims to change the bash completion to correctly suggest and complete commands for the updated CLI.


agrover:
LGTM, merging. Thanks!

PR #221, opened 2018-12-14
Move an unused constant

mulkieran:
Remove MANAGER_INTERFACE from _constants.py, as it is unused. Rename it to
_MANAGER_INTERFACE and define it in _data.py, where it can be used.


agrover:
trivial, merging.

PR #223, opened 2018-12-17
Add a summary message for engine errors

mulkieran:
This will a help a user of the CLI understand the context of the error a bit
better.

### DIFF
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -26,6 +26,7 @@ from dbus_client_gen import DbusClientUniqueResultError
 from dbus_client_gen import DbusClientUnknownSearchPropertiesError
 
 from ._actions import interface_name_to_common_name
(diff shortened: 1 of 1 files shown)


mulkieran:
The effect looks something like this, where the engine message is completely faked up in the stratisd sim engine for demonstration purposes.

```
[root@localhost stratis-cli]# PYTHONPATH=./src ./bin/stratis pool create name /dev/dev1
Execution failure caused by:
ERROR: Device has an existing signature. Use wipefs, etc.

stratisd failed to perform the operation that you requested. It returned the following information via the D-Bus: ERROR: Device has an existing signature. Use wipefs, etc..
[root@localhost stratis-cli]# 
```

mulkieran:
I don't want this to be merged today, since it is a bit new.

mulkieran:
I would like to take a short time to consider if the EngineError message ought to be enhanced.

PR #226, opened 2019-01-10
Adjust stratisd version check to allow snapshot releases

agrover:
If we call the version parts x.y.z, let's allow the cli to work with all
1.0.z releases, since we don't want to have to update this value every
time we do a snapshot (where z is incremented). I just tagged a 1.0.3,
in fact, which we want to not trigger the version check.

The consequence is that when we *do* land any API changes in stratisd, we
will need bump the y version. (Maybe at some point we should break API
version out from stratisd's version, but for now they are linked.)

### DIFF
--- a/src/stratis_cli/_actions/_data.py
+++ b/src/stratis_cli/_actions/_data.py
@@ -29,7 +29,7 @@ from .._errors import StratisCliValueError
 
 from ._constants import DBUS_TIMEOUT_SECONDS
 
-MAXIMUM_STRATISD_VERSION = (1, 0, 2)
+MAXIMUM_STRATISD_VERSION = (1, 0, 99)
 MINIMUM_STRATISD_VERSION = (1, 0, 0)
 
 # a sanity check on the relation between the minimum and maximum versions


mulkieran:
> We currently aren't adhering to semantic versioning, so doing any kind of stringent version checking seems moot, as does having a discussion on the names of variables pertaining to it. All of our CI is failing because of this, so I believe we should get this in, tag new client release so we can move forward.

I agree that it would be desirable to get Jenkins stratis-cli tests to
no longer fail on version checking. But I don't think it's desirable to
merge an obvious bug.

The bug is that all the code that interacts with the constant named
MAXIMUM_STRATISD_VERSION treats the value as a maximum. The change to
the value changes its meaning also, it is now a limit, not a maximum.
So, all the code that treats it as a maximum now works incorrectly,
including, but not limited to, the constant name, which communicates the
wrong information to the reader.

There are several alternatives:

1. Revert the previous commits which introduced the version check.

2. Fix up this PR so all the code follows the intent of the PR, that is,
to substitute a limit for a maximum.

3. Fix up this PR so that the maximum value matches the current stratisd
version value, i.e., stick to the meaning of maximum.

4. 1 followed by 2.

There should be no need to make a new stratis-cli version. The Jenkins
stratis-cli tests pull from the HEAD of the current master, so once any
of these suggested changes go in the tests will be no longer be failing
due to the version check.

agrover:
Merging to fix breakage.

Issue #205, opened 2018-10-10
Column alignment  fails with non-ascii characters


tasleson:
```
# stratis filesystem
Pool Name    Name  Used     Created            Device                       
yes_you_can  ☺     546 MiB  Oct 05 2018 16:24  /dev/stratis/yes_you_can/☺   
yes_you_can  漢字    546 MiB  Oct 10 2018 09:37  /dev/stratis/yes_you_can/漢字
```

tasleson:
I tried experimenting with `wcwidth` and the widths seem to be calculated correctly, however things do not align because the characters are not 1 column width wide.  This seems to be more of a font issue than a calculating widths issue.  Not sure if we can fix this one.

``` diff
--- a/src/stratis_cli/_actions/_formatting.py
+++ b/src/stratis_cli/_actions/_formatting.py
@@ -16,6 +16,7 @@ Formatting for tables.
 """
 
 import sys
+from wcwidth import wcswidth
 
 
 def print_table(column_headings, row_entries, alignment, file=sys.stdout):
@@ -38,7 +39,7 @@ def print_table(column_headings, row_entries, alignment, file=sys.stdout):
 
     column_lengths = [
         max(
-            max((len(e[index]) for e in row_entries), default=0),
+            max((wcswidth(e[index]) for e in row_entries), default=0),
             len(column_headings[index])) + 2 for index in range(num_columns)
     ]

```

cathay4t:
@tasleson Have you try https://docs.rs/ucd/0.1.1/ucd/tables/misc/enum.EastAsianWidth.html ?

cathay4t:
And this is how unicode standard defines it https://www.unicode.org/reports/tr11/

mulkieran:
This seems like:

* Not a fundamental correctness issue.
* Kind of challenging to fix.

Can we postpone it?

mulkieran:
The rule has been that the PRs get put in the "Review" slot, but issue are supposed to remain in "In Progress" until they clear.

tasleson:
> The rule has been that the PRs get put in the "Review" slot, but issue are supposed to remain in "In Progress" until they clear.

Indeed, I've corrected things, thanks.

mulkieran:
Resolved by #245.

PR #230, opened 2019-02-04
Correct column alignment for unicode characters

tasleson:
When we calculate the column widths we need to use how many columns
the text will occupy, not the number of characters.  When we output
the text for each piece of the column we need to adjust the width
to compenstate for this as well.

Corrects: https://github.com/stratis-storage/stratis-cli/issues/205


tasleson:
Replaced with https://github.com/stratis-storage/stratis-cli/pull/234

PR #232, opened 2019-02-05
Revert dynamic version checking

mulkieran:
Do not dynamically check for the stratisd version when running the stratis CLI.


mulkieran:
I don't think our release process is adequate to support this, and it's a nice idea, but it hasn't shown its utility yet. As I pointed out in https://github.com/stratis-storage/stratis-cli/pull/207#discussion_r234236547, the problem is maintaining the minimum and maximum version correctly. We haven't really done so well with this. Let's get rid of it before we release it.

mulkieran:
We should make sure of our plan for new releases before the "do not merge" flag is removed.

PR #233, opened 2019-02-06
Tasleson column align

mulkieran:
Still working.

### DIFF
--- a/src/stratis_cli/_actions/_formatting.py
+++ b/src/stratis_cli/_actions/_formatting.py
@@ -17,6 +17,11 @@ Formatting for tables.
 
 import sys
 
+try:
+    from wcwidth import wcswidth
+except ImportError:
+    wcswidth = len
+
 
 def print_table(column_headings, row_entries, alignment, file=sys.stdout):
     """
@@ -38,8 +43,9 @@ def print_table(column_headings, row_entries, alignment, file=sys.stdout):
 
     column_lengths = [
         max(
-            max((len(e[index]) for e in row_entries), default=0),
-            len(column_headings[index])) + 2 for index in range(num_columns)
+            max((wcswidth(e[index]) for e in row_entries), default=0),
+            wcswidth(column_headings[index])) + 2
+        for index in range(num_columns)
     ]
 
     for index in range(num_columns):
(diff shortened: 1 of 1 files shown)


src/stratis_cli/_actions/_formatting.py
```diff
@@ -52,9 +58,15 @@ def print_table(column_headings, row_entries, alignment, file=sys.stdout):
 
     for row in row_entries:
         for index in range(num_columns):
+            calc_w = column_lengths[index]
+            entry = row[index]
+
+            # -1 is returned for non-printable wide character
```
mulkieran: This part is puzzling.

tasleson: The only time it's a no op is when we don't have `wcwidth` package installed.  I did this because I wasn't able to find the package for the version of ubuntu running on travis.

`wcswidth` will either return the same value as `len` or something slightly bigger if one or more characters occupy more than one column width, thus we need to decrease the size of width by the difference to get the columns to align when the characters are printed to the terminal.

mulkieran: Slightly smaller is also possible:

```
>>> wcswidth(u'\u0065\u0301')
1
```

tasleson: > Slightly smaller is also possible

Yes, this is possible, but it is also handled as it then subtracts a negative number which causes us to increase the width the corresponding amount which causes things to align correctly too.

eg.
```
# /root/stratis fs list
Pool Name  Name  Used     Created            Device                 UUID                              
unicode    e     546 MiB  Feb 07 2019 15:33  /stratis/unicode/e     3bf22806a6df4660aa527d646209595f  
unicode    eeee  546 MiB  Feb 07 2019 15:33  /stratis/unicode/eeee  17101e39e72e423c90d8be5cb37c055b  
unicodé    é     546 MiB  Feb 07 2019 15:33  /stratis/unicodé/é     0c2caf641dde41beb40bed6911f75c74  
unicodé    éééé  546 MiB  Feb 07 2019 15:33  /stratis/unicodé/éééé  4ecacb15fb64453191d7da731c5f1601
```

tasleson: As we currently have in master

```
# /root/stratis fs list
Pool Name  Name      Used     Created            Device                      UUID                              
unicode    e         546 MiB  Feb 07 2019 15:33  /stratis/unicode/e          3bf22806a6df4660aa527d646209595f  
unicode    eeee      546 MiB  Feb 07 2019 15:33  /stratis/unicode/eeee       17101e39e72e423c90d8be5cb37c055b  
unicodé   é        546 MiB  Feb 07 2019 15:33  /stratis/unicodé/é        0c2caf641dde41beb40bed6911f75c74  
unicodé   éééé  546 MiB  Feb 07 2019 15:33  /stratis/unicodé/éééé  4ecacb15fb64453191d7da731c5f1601
```

PR #235, opened 2019-02-07
Tasleson column align final

mulkieran:
An explanation of the overall purpose of the suggested changes.

For this, first I want to define what the PR promises. In master, columns are aligned correctly if all entries, both header and data, are standard width characters. This PR makes this statement true, given the broader condition that the characters may be any printable characters, standard width or not. It does not guarantee to align the columns correctly if an entry contains unprintable characters. It seems to me that ensuring the alignment is correct if entries contain all possible characters, even unprintable ones, is not worth doing.

A first step is to make wcwidth dependency unconditional. It makes the code simpler, and if there is no reason not to, then it might as well be done.

The second step is to assert that the method does not promise correctness in case an entry contains unprintable characters. As I have stated above, I don't think this extreme correctness goal is worth bothering with. It removes a check for unprintable characters in one place. This check did nothing to make the method more correct, since an entry with an unprintable character might otherwise have many 2 or zero width characters. It was also a source of inconsistency, because this '-1' value was not checked for when calculating the column width in the first place.

It makes the method more correct, by also ensuring that any header items that might contain non-standard width characters are printed out correctly. This step also makes the code much simpler since printing of header and data rows becomes identical and so can be abstracted.

It turns out that the necessary adjustment to make in the number of characters passed to the format method in order to get correct alignment is hard to understand. This PR abstracts that crucial operation into a separate method and tries to explain why it is correct a bit.

Also, some simple tests that allow a visual check of correctness are added.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,10 +36,8 @@ setuptools.setup(
         'Topic :: System :: Filesystems', 'Topic :: Systems Administration'
     ],
     install_requires=[
-        'dbus-client-gen>=0.4',
-        'dbus-python-client-gen>=0.5',
-        'justbytes==0.11',
-        'python-dateutil',
+        'dbus-client-gen>=0.4', 'dbus-python-client-gen>=0.5',
+        'justbytes==0.11', 'python-dateutil', 'wcwidth'
     ],
     package_dir={"": "src"},
     packages=setuptools.find_packages("src"),
--- a/src/stratis_cli/_actions/_formatting.py
+++ b/src/stratis_cli/_actions/_formatting.py
@@ -35,7 +35,7 @@ def _get_column_width_chars(column_width_cells, entry):
 
     :returns: the column width in characters
 
-    Precondition: mcswidth(entry) != -1
+    Precondition: wcswidth(entry) != -1
                   (equivalently, entry has no unprintable characters)
     """
     return column_width_cells - (wcswidth(entry) - len(entry))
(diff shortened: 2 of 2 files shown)


mulkieran:
Only proven correct, not tested. How to test?

mulkieran:
Record some strategies that weren't wroth using in final version.

```
    # Import hypothesis in main method so src does not depend on
    # hypothesis package.
    from hypothesis import strategies

    @strategies.composite
    def table_strategy(draw):
        """
        Strategy for generating tables with arbitrary characters.

        FIXME: Generates a few characters that can not be printed
        on my terminal.
        """

        num_columns = draw(strategies.integers(min_value=1, max_value=5))
        num_rows = draw(strategies.integers(min_value=1, max_value=5))
        table = draw(
            strategies.lists(
                strategies.lists(
                    strategies.text(
                        strategies.characters(
                            # Use only characters in BMP 0
                            max_codepoint=65535,
                            blacklist_categories=('Cc', 'Cf', 'Co', 'Cs', 'Sk',
                                                  'Sm', 'So', 'Zl', 'Zp',
                                                  'Zs')),
                        min_size=1,
                        max_size=10),
                    min_size=num_columns,
                    max_size=num_columns),
                min_size=num_rows,
                max_size=num_rows))
        return table

    def alignment_strategy(num_alignments):
        """
        A strategy to generate alignments.
        """
        return strategies.lists(
            strategies.sampled_from(['<', '>', '^']),
            min_size=num_alignments,
            max_size=num_alignments)
```

mulkieran:
Changes merged into #245 .

Issue #237, opened 2019-02-13
Update manpage for cache size restriction


mulkieran:
https://bugzilla.redhat.com/show_bug.cgi?id=1676801

mulkieran:
Resolved by #239.

PR #238, opened 2019-02-13
Update manpage to indicate temporary restriction on cache size

mulkieran:
Resolves: #237.

### DIFF
--- a/docs/stratis.txt
+++ b/docs/stratis.txt
@@ -49,6 +49,10 @@ pool add-cache <pool_name> <blockdev> [<blockdev>..]::
 	 Add one or more blockdevs to an existing pool, to be used as a cache
 	 instead of additional storage. Typically, smaller and faster drives,
 	 such as SSDs, are used for this purpose.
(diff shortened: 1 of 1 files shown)


docs/stratis.txt
```diff
@@ -49,6 +49,10 @@ pool add-cache <pool_name> <blockdev> [<blockdev>..]::
 	 Add one or more blockdevs to an existing pool, to be used as a cache
 	 instead of additional storage. Typically, smaller and faster drives,
 	 such as SSDs, are used for this purpose.
+         NOTE: There is a temporary restriction on the total size of the
+         cache device of 32 TiB. Adding devices to the cache so that the
+         cumulative size of all the devices in the cache exceeds 32 TiB will
+         result in the engine returning an error.
```
tasleson: I don't think command line users should care about knowing that there is an `engine` and that it's returning as error.  IMHO simply stating `... will result in an error.` would be fine.

mulkieran: Sure.

mulkieran:
Superceded by #239.

PR #242, opened 2019-02-16
Tidy up Travis configuration

mulkieran:
* Leave out dist specification; trusty became the default years ago
* Use the apt configuration specification instead of explict apt instructions.
It is the currently preferred way.
* Specify different versions of Python for lint and fmt tasks, and explain
the choice of the Python versions in comments.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,10 @@
 os: linux
-dist: trusty
-
-before_install:
-    - sudo apt-get -qq update
-    - sudo apt-get install -y libdbus-glib-1-dev
-
 language: python
-python: "3.4"
+
+addons:
+  apt:
+    packages:
+      libdbus-glib-1-dev
 
 branches:
     only: master
@@ -14,6 +12,8 @@ branches:
 matrix:
     include:
         - env: TASK=lint
+          python: "3.4" # dbus-python does not compile on higher versions
         - env: TASK=fmt-travis
+          python: "3.6" # Run fmt task on newest version currently available
 
 script: make -f Makefile $TASK


.travis.yml
```diff
@@ -1,19 +1,19 @@
...
+    packages:
+      libdbus-glib-1-dev
 
 branches:
     only: master
 
 matrix:
     include:
         - env: TASK=lint
+          python: "3.4" # dbus-python does not compile on higher versions
```
tasleson: Please elaborate on this.  AFAIK `dbus-python` appears to work fine on my python 3.7.2 system?

mulkieran: It fails on Travis, which is running Ubuntu. You can look at examples of how it fails here:
https://travis-ci.org/stratis-storage/into-dbus-python/builds/494037710.

mulkieran: Note that this is the same decision, here: https://github.com/stratis-storage/stratisd/pull/1436.

tasleson: Is there some reason we can't use a package install instead of building from source?  All we are trying to do is lint the code.

mulkieran: What do you mean by a package install?

tasleson: > What do you mean by a package install?

I mean satisfy a python library dependency by using a package install ref. 
https://packages.ubuntu.com/trusty/python3-dbus
instead of building dbus-python from source.

You are using a package install to satisfy the dependency of `libdbus-glib-1-dev`

mulkieran: Yes. But from that link you gave me:

```
python3 (<< 3.5)
    interactive high-level object-oriented language (default python3 version)
```

So, I think it's the same either way.

mulkieran:
@tasleson I've opened an issue about moving Travis distribution up to xenial: https://github.com/stratis-storage/stratis-cli/issues/244, but at this time, I think it makes sense to merge this PR.

PR #249, opened 2019-03-08
Catch KeyboardInterrupt at the outermost level

mulkieran:
Closes: #250.

### DIFF
--- a/bin/stratis
+++ b/bin/stratis
@@ -26,7 +26,10 @@ def main():
     """
     The main method.
     """
-    return run()(sys.argv[1:])
+    try:
+        return run()(sys.argv[1:])
+    except KeyboardInterrupt:
+        sys.exit("Received KeyboardInterrupt")
 
 
 if __name__ == "__main__":


mulkieran:
This derives from a bz. Since it is, AFAICT, so straightforward to satisfy, I think it might as well be taken care of now.

mulkieran:
> In the BZ https://bugzilla.redhat.com/show_bug.cgi?id=1686652#c1 you mention returning an exit code of 130. This change results in an exit of 1 with the message going to stderr. I'm thinking that it would be good for someone scripting the command line to know if it got interrupted by a signal with the command line returning a unique code. This would allow the caller to know that the request will likely complete as requested at some point in the future.

It is easy enough to return an exit code of 130.

I chose to go w/ 1, because it is correct, if not very precise.

I think it makes sense to move this ahead as is, and file an issue about working out status codes. It's one of those things that is surprisingly lacking in standards and apparently some user would object to any status code value above 127.

tasleson:
Follow-up issue https://github.com/stratis-storage/stratis-cli/issues/251

mulkieran:
Superceded by #254.

PR #253, opened 2019-03-13
Integration black box tests

tasleson:
Set of unit test to verify basic functionality of `stratis` CLI and `stratisd` from the point of a packaged install.


tests/blackbox/stratis_cert.py
```diff
@@ -0,0 +1,347 @@
...
+        if result[0]:
+            stdout_text = bytes(result[0]).decode("utf-8")
+        if result[1]:
+            stderr_text = bytes(result[1]).decode("utf-8")
+
+        self.assertTrue("Traceback" not in stdout_text)
+        self.assertTrue("Traceback" not in stderr_text)
+        self.assertNotEqual(process.returncode, 0)
+
+    @unittest.expectedFailure
```
mulkieran: Please change this test. It hasn't been agreed that putting the name in the error is the appropriate thing, see comment: https://bugzilla.redhat.com/show_bug.cgi?id=1686852#c1.

requirements.txt
```diff
@@ -1,2 +1,3 @@
+rpm
```
mulkieran: I don't think this is necessary unless the tests are being run as part of Travis, which isn't happening now, AFAICT.

tests/blackbox/stratis_cert.py
```diff
@@ -0,0 +1,347 @@
...
+        StratisCli.pool_create(pool_name, [DISKS[0]])
+        StratisCli.fs_create(pool_name, fs_name)
+        StratisCli.fs_rename(pool_name, fs_name, fs_new_name)
+
+        fs = StratisCli.fs_list()
+        self.assertTrue(fs_new_name in fs.keys())
+        self.assertFalse(fs_name in fs.keys())
+        self.assertEqual(1, len(fs))
+
+    @unittest.expectedFailure
```
mulkieran: This test should succeed, since the fix has been merged, so expectedFailure should be taken off.

tests/blackbox/stratis_cert.py
tasleson: Correct no determination has been made, thus we can leave it as is because we are expecting it to fail.  When we expect it to pass we can then make the final changes to what it should be checking for.

requirements.txt
tasleson: Not necessary, but is it hurting anything?

tests/blackbox/stratis_cert.py
tasleson: Until the correction gets packaged into a release this need to stay as is.

tests/blackbox/stratis_cert.py
mulkieran: I think it is better that the test test for the most important thing, which is that if the pool is created with disks that have already been claimed by another pool, then the command will return an exit code of 1. Because this test is marked as an expected failure, if that becomes not true, the test will still succeed. The test is most valuable if the ```expectedFailure``` and the following lines are stripped out. 
```
        # Users want the pool name in the error output
        self.assertTrue(pool_name in stdout)
```
and is left as a normal running test.

Checks for what the error messages actually contains can be added later to refine the test, if that seems like the correct thing to do.

tests/blackbox/stratis_cert.py
mulkieran: Oh, right.

tests/blackbox/stratis_cert.py
tasleson: All of the tests that are decorated with `@unittest.expectedFailure` are for regression tests for existing bugs.  The sole purpose of this test is to ensure that the pool name is in the output.  In my opinion the end user is correct that the pool name should be included in the error message.  I believe this is the correct thing to do because everything in the CLI is name based.  Users never specify anything by UUID, never see anything by UUID, thus it makes sense that the error message would specify the pool by name.  Having both the UUID and name in the output would be fine in my opinion too.

tasleson:
I'll add a readme to the blackbox directory

requirements.txt
mulkieran: I'm only able to imagine two possibilities: either it is useful or it is not.

Being puzzled by its occurrence, I looked over the PR for a bit, trying to figure out how it was used and I concluded that it didn't seemed to be used at all, at least in this code base.

Assuming it is not used, it makes no sense to me for it to be merged. If it is used in some context outside this code base, it makes sense to document what that use is so that some person doesn't notice it's not used and remove it during some later infrastructure update, thereby breaking some code outside this code base that somehow relies on it.

tasleson: `tests/blackbox/testlib/utils.py`
``` python
import rpm
...
def rpm_package_version(name):
    """
    Retrieve the version of the specified package
    :param name: Name of package
    :return: String representation of version, None if package not found.
    """
    ts = rpm.TransactionSet()
    mi = ts.dbMatch('name', name)

    for i in mi:
        return i["version"].decode("utf-8")

    return None
```

tasleson:
@mulkieran @trgill If it would be less confusing for people, we can move these tests to the packaging repo where they will be run.  I initially wanted to have them here so that anyone could see them and improve them, but the primary reason for existence is to test rpm packaged releases, so they could exist there instead.

tasleson:
Replaced with https://github.com/stratis-storage/stratis-cli/pull/257

PR #261, opened 2019-04-23
Move CI scripts out of Jenkins and into the git repo

trgill:
The script creates a STRATIS_DEPS_DIR at the top level to hold the needed
dependencies of the tests.

The STRATISD environment variable is set to the stratisd executable.  The
python code starts/stops the process as necessary for the tests.

### DIFF
--- a/tests/ci/cli.sh
+++ b/tests/ci/cli.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+set -e
+
+# Jenkins will set WORKSPACE to the top level directory that contains
+# the stratis-cli git repo
+if [ -z "$WORKSPACE" ]
+then
+    echo "Required WORKSPACE environment variable not set"
+    exit 1
+fi
+
+export STRATIS_DEPS_DIR=$WORKSPACE/stratis-deps
+
+cd $WORKSPACE
+
+# If there is a stale STRATIS_DEPS_DIR remove it
+if [ -d $STRATIS_DEPS_DIR ]
+then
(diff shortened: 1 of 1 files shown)


tests/ci/cli.sh
```diff
@@ -0,0 +1,51 @@
...
+mkdir $STRATIS_DEPS_DIR
+cd $STRATIS_DEPS_DIR
+
+# Clone the dependencies
+git clone https://github.com/stratis-storage/dbus-python-client-gen.git
+git clone https://github.com/stratis-storage/dbus-client-gen.git
+git clone https://github.com/stratis-storage/into-dbus-python.git
+git clone https://github.com/stratis-storage/dbus-signature-pyparsing.git
+git clone https://github.com/stratis-storage/stratisd.git
+
```
bgurney-rh: After the stratisd repository is cloned, it's probably good to add the stratisd.conf to dbus:
```
cp stratis-deps/stratisd/stratisd.conf /etc/dbus-1/system.d/
```
Otherwise, the cli.sh test will fail with "Could not get owner of name 'org.storage.stratis1': no such name".

bgurney-rh: However, if the stratisd.conf is copied in prior to the script execution, the tests all succeed on my test system.

trgill: Thanks.   I agree the script should make it obvious what is wrong.  We may want to make the tests do the same thing.

I'll also update the stratisd script.  It has a similar issue.

PR #262, opened 2019-04-26
Add path to .cargo/bin to include the "cargo" command

trgill:
Jenkins uses the fedora-user ID to connect to cloud nodes.  The script is
invoked via "sudo -E ./tests/ci/cli.sh" because it requires root
permissions.  The root shell by default does not have a path to cargo set.

Signed-off-by: Todd Gill <tgill@redhat.com>

### DIFF
--- a/tests/ci/cli.sh
+++ b/tests/ci/cli.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -e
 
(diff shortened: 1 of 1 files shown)


trgill:
@mulkieran this fixes a mistake I made in the previous PR.  With this change the Jenkins task is just:

sudo -E ./tests/ci/cli.sh

PR #269, opened 2019-05-22
Change display on error

mulkieran:
Previously, if the action failed and an exception was raised, an abbreviated version of the chained exceptions was displayed to the user. If there was a summary explanation for the problem, indicating the most likely cause, that was displayed at the end.

Now, if there is a summary explanation, only the summary explanation is displayed. If there is no explanation, then it is considered an unexpected error and a completely different action is taken:

1. A message is printed indicating that the error is unexpected and that the user should report the error at the stratis-cli GitHub site.
2. The exception is reraised.

The result of this is that what the user sees is the error message described in (1) followed by a display of the stack trace, including all chained exceptions.

There are only two drawbacks to this approach that I can see:
* It might be that the explanation is wrong. In order to avoid this, the explanations are constructed to say only what is actually certain, but this could still be misleading. Even a carefully phrased explanation might be misinterpreted by a user to be more definite than it actually is. If the explanation is really wrong, then a process is required to find out what the true cause actually is, and no stack trace is really available for diagnosis. However, at least some of the stack trace can be inferred from the code and the explanation, so that would not leave us totally at a loss.
* The simplest and best thing to do, in order to have the stack trace properly printed, is to reraise the exception and allow the Python interpreter to handle displaying it. This means that in this case it is the Python interpreter that chooses the error code for the process, not stratis-cli.

Also, please note that there is one known error for which there is no summary explanation. That is the one where GetManagedObjects() fails because some property cannot be obtained by stratisd. I think that in this case it is better to fix the GetManagedObjects failure on the stratisd side, but if necessary, the exception could be fielded and explained.

### DIFF
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -122,9 +122,8 @@ def handle_error(err):
     if explanation is None:
         exit_msg = (
             "stratis encountered an unexpected error during execution. "
-            "Please report the error at "
-            "https://github.com/stratis-storage/stratis-cli, and "
-            "include the stack trace below.")
+            "Please report the error and include in your report the stack "
+            "trace shown below.")
         print(exit_msg, os.linesep, file=sys.stderr, flush=True)
         raise err
 
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -50,19 +50,6 @@ def get_errors(exc):
             return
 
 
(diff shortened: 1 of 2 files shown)


mulkieran:
Please see #269, for examples of the output in the two different error cases.

mulkieran:
I'm pulling this back to in progress. We might want some wording changes, and I think we should change how it handles AttributeErrors as well.

mulkieran:
Superceded by #274 , closing.

PR #272, opened 2019-05-23
blackbox: fix units_to_bytes return

bgurney-rh:
The units_to_bytes function is always failing due to the input
variable "size" always containing a period (e.g.: "10.00"); the
call fails with "ValueError: invalid literal for int() with base 10".

To avoid losing precision on the size calculation, change the call
to "float(size)", which should always work when there is a period
in the string.

### DIFF
--- a/tests/blackbox/testlib/utils.py
+++ b/tests/blackbox/testlib/utils.py
@@ -49,7 +49,7 @@ def units_to_bytes(size, units):
         "EiB": 1024**6,
         "ZiB": 1024**7
     }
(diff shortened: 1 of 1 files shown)


bgurney-rh:
This would run into problems with precision due to floating point error, as well as the way that real block devices could have a non-power-of-two sector count for various reasons (resulting in a "size" input valie of "10.00" instead of just "10".)

However, it turns out that the return value from units_to_bytes() is only really used as a string, so there's an easier fix that achieves the same end result.  That will be in a future pull request.

PR #275, opened 2019-05-24
blackbox: make units_to_bytes return a string

bgurney-rh:
The units_to_bytes function is always failing due to the input
variable "size" always containing a period (e.g.: "10.00"); the
call fails with "ValueError: invalid literal for int() with base 10".

However, all of the callers of units_to_bytes() only use the return
value as a string in a dictionary, therefore returning the value
as a concatenation of size + units allows the tests to pass.

### DIFF
--- a/tests/blackbox/testlib/stratis.py
+++ b/tests/blackbox/testlib/stratis.py
@@ -18,10 +18,7 @@ import datetime
 import os
 import time
 
-from .utils import (exec_command,
-                    rs,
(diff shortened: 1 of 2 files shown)


bgurney-rh:
Also, change the units_to_bytes() function name to size_representation(), since this function now returns a string representation of the size.

bgurney-rh:
Rebased.  The travis-ci test that was failing now passes.

PR #295, opened 2019-07-30
Remove a few unnecessary pylint directives

mulkieran:
This sweep removes a bunch of unnecessary pylint directives. All remaining directives are required.


mulkieran:
I happened to notice an unnecessary directive while looking for examples of Python error-management...then I noticed just a few more.

PR #296, opened 2019-07-30
Make bash completion aware of --propagate

carzacc:
The bash completion script was previously unaware of the existence of `--propagate` at all.

With this PR it
* autocompletes the word `--propagate` anywhere in the command;
* ignores the presence of `--propagate` when looking for the previous arguments and counting the number of arguments, so that it doesn't mistake it for the previous argument the user is supposed to provide to the `stratis` command.

Signed-off-by: Carmine Zaccagnino <carmine@carminezacc.com>

### DIFF
--- a/shell-completion/bash/stratis
+++ b/shell-completion/bash/stratis
@@ -1,15 +1,33 @@
 _stratis()
 {
-	local cur prev x2prev x3prev opts root_subcommands pool_subcommands fs_subcommands blockdev_subcommands daemon_subcommands complete_pools blockdevs first second third
+	local cur prev x2prev x3prev opts root_subcommands pool_subcommands fs_subcommands blockdev_subcommands daemon_subcommands complete_pools blockdevs first second third comp_cword_adj
+
 	COMPREPLY=()
+	comp_words_adj=()
+
+	comp_cword_adj="${COMP_CWORD}"
+
+# Turn comp_cword_adj and comp_word_adj into versions of COMP_WORD
+# and COMP_CWORD that pretend --propagate doesn't exist
+	for word in "${COMP_WORDS[@]}"
+	do
+		if [[ ${word} == "--propagate" ]]; then
+			((comp_cword_adj--))
+		else
+			comp_words_adj+=(${word})
+		fi
+	done
+	
+# The current word needs to be able to be completed for --propagate
 	cur="${COMP_WORDS[COMP_CWORD]}"
-	prev="${COMP_WORDS[COMP_CWORD-1]}"
-	x2prev="${COMP_WORDS[COMP_CWORD-2]}"
-	x3prev="${COMP_WORDS[COMP_CWORD-3]}"
-	first="${COMP_WORDS[1]}"
-	second="${COMP_WORDS[2]}"
-	third="${COMP_WORDS[3]}"
-	opts="-h --help --version"
+# The other words must ignore --propagate and use adjusted values
+	prev="${comp_words_adj[comp_cword_adj-1]}"
+	x2prev="${comp_words_adj[comp_cword_adj-2]}"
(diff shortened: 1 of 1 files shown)


mulkieran:
ok to test

poizen18:
Hello Carmine,

Nice find that it doesn't see --propagate 
> The bash completion script was previously unaware of the existence of `--propagate` at all.
> 
> With this PR it
> 
>     * autocompletes the word `--propagate` anywhere in the command;
--propogate only needs to be completed in the beginning of the command. I say this because the man page suggests : 

 ```
SYNOPSIS
       stratis [GLOBAL OPTIONS] pool <command> `
 :::
GLOBAL OPTIONS
       --version
           Show stratis-cli version.

       --help, -h
           Show help on command.

       --propagate
           (For debugging.) Allow exceptions raised during execution to
           propagate.
```

>     * ignores the presence of `--propagate` when looking for the previous arguments and counting the number of arguments, so that it doesn't mistake it for the previous argument the user is supposed to provide to the `stratis` command.

Would you think the following patch makes sense or something similar? 


```
--- a/shell-completion/bash/stratis
+++ b/shell-completion/bash/stratis
@@ -1,3 +1,5 @@
+#!/bin/bash -x 
+
 _stratis()
 {
        local cur prev x2prev x3prev opts root_subcommands pool_subcommands fs_subcommands blockdev_subcommands daemon_subcommands complete_pools blockdevs first second third
@@ -9,8 +11,8 @@ _stratis()
        first="${COMP_WORDS[1]}"
        second="${COMP_WORDS[2]}"
        third="${COMP_WORDS[3]}"
-       opts="-h --help --version"
-       root_subcommands="--help --version daemon pool blockdev filesystem"
+       opts="-h --help --version --propagate"
+       root_subcommands="--propagate --help --version daemon pool blockdev filesystem"

```

**Comment by carzacc** [comment]
Yes, @poizen18, autocompleting it as a root subcommand and not as an option might work if it is undesirable to have the autocomplete suggest `--propagate` anywhere in the command. The same could be done with `--version` actually, so that it doesn't get suggested later in the command as well.

**Comment by carzacc** [comment]
I've taken another look at the code and it needs more changes than just moving `--propagate` and `--version` from the `opts` into the `root_subcommands`: we need to create a separate `global_opts` variable and change the code that handles the `opts` by adding a nested `if` that checks whether we're at the root, adding the global options to the `COMPREPLY` if that is the case.
I might not have expressed myself in the best and clearest way possible in this comment, I will fix the issue in the code and it will be clear.

**Comment by mulkieran** [comment]
Thanks for taking a look @poizen18. Do you have any further comments?

**Comment by poizen18** [comment]
The new version does not complete for `--propagate` when pressing <tab><tab> in the beginning.

>Make bash completion aware of --propagate
If this is the only reason, we can simply add --propagate in globalopts variable and use it with `COMPREPLY`.

I am sorry, could you help me understand why the following would be necessary?

```
# Turn comp_cword_adj and comp_word_adj into versions of COMP_WORD
# and COMP_CWORD that pretend --propagate doesn't exist
```
Instead, we can show the user/admin that --propagate is an option only in the beginning and then forget about it. This can be done by just adding globalopts and calling it. Please let me know if I have misunderstood the purpose of your patch. Sorry if I have done that. 

Awaiting your feedback @carzacc

carzacc:
@poizen18 that's what makes the completion ignore `--propagate`. Since it's only in the beginning we can simplify everything and that's not needed anymore, I'll remove it.

poizen18:
@carzacc  after applying `df2664a` ; bash completion isn't working any more.

Do you want to help me rewrite the code so that we do not have to write hacks around --propagate or do you want to modify existing code?

carzacc:
It's working on my system, what exactly isn't working anymore?

carzacc:
OK, I'll do that now.

PR #300, opened 2019-08-07
Make bash completion aware of --propagate

carzacc:


mulkieran:
ok to test

mulkieran:
@carzacc There is a slight difference between the code changes in this and the previous PR. Did you add some additional fix?

bgurney-rh:
Reading back to the comments of #296, the following section (around line 120 in the new version of shell-completion/bash/stratis) is retaining "COMP_CWORD" (which is why it doesn't appear in this diff):

```
       Code to handle multiple block device names or multiple filesystem names for filesystem destroy and pool create/add-data/add-cache
               if [[ ${COMP_CWORD} > 3 ]]; then
                       case ${first} in 
                               filesystem|fs)
                                       case ${second} in 
```

carzacc:
@bgurney-rh @mulkieran `comp_cword_adj` doesn't exist anymore because it's part of the changes that were rolled back after @poizen18's comments. That part should retain the use of `COMP_CWORD`, as it does in this PR, and leaving it there in the previous PR was a blunder on my part while implementing the changes requested by @poizen18.

PR #294, opened 2019-07-24
Master remove properties

mulkieran:

Issue #292, opened 2019-06-12
RFE: Create machine-readable output option for "stratis ... list" commands


bgurney-rh:
It would be nice for the "stratis ... list" commands to have a "machine-readable" output option (e.g.: YAML, JSON, TOML, etc.), to allow other programs to interpret the results without resorting to "screen-scraping".

Example output from a stratis pool created on a test device:

```
# stratis pool create spool1 /dev/nvme0n1p5
# stratis fs create spool1 sfs1
# stratis fs create spool1 sfs2
# stratis fs create spool1 sfs3
# stratis pool list
Name    Total Physical Size  Total Physical Used
spool1           228.83 GiB             1.85 GiB
# stratis fs list
Pool Name  Name  Used     Created            Device                UUID
spool1     sfs1  545 MiB  Jun 12 2019 15:46  /stratis/spool1/sfs1  5e4244e98f814df1be8247430ab3e741
spool1     sfs2  545 MiB  Jun 12 2019 15:46  /stratis/spool1/sfs2  446cd0f3b08a4a7f976774427d664b4e
spool1     sfs3  545 MiB  Jun 12 2019 15:46  /stratis/spool1/sfs3  f9dc03a10e694c6b94024512f835af7a
# stratis blockdev list
Pool Name  Device Node     Physical Size  State  Tier
spool1     /dev/nvme0n1p5     228.83 GiB  InUse  Data
```

mulkieran:
We would like to avoid having anybody commit to parsing the output from the CLI as it stands now w/ anything but their human eyes, that's for sure. One, among many, excellent reasons for this is that the sizes displayed in general do not represent exact values, but a range of values containing a finite, but potentially quite large, set of elements.

We're in a different position from a lot of other apps, because the CLI talks to the daemon across the D-Bus. In particular, the information about sizes that is transmitted across the D-Bus is _exact_, _unlike_ that which is displayed in the CLI (which for formatting reasons is generally inexact).

We should anticipate specifying a structured output mode, and responding w/ structured output for _every_ command, not just the list commands, in order to maintain consistency. We should also expect that the structured output should be richer and more exact than the human-reader directed output, since it is unconstrained by the formatting issues that plague human readable output.

The _implementation_ of this should be relatively easy; the _design_ of the structured output is the more challenging part. I would like it to be able to be formally validatable, if at at all possible.

mulkieran:
How should we request machine readable output? I like the environment variable approach.

[...]

PR #301, opened 2019-08-10
Disentangle formatting of list output from getting objects

mulkieran:
The basic contribution of this PR is to untangle getting information related to formatting from the action of getting objects to format.

This is necessary, because we need three steps for completing #294 and #292 
1. Getting the objects to display.
2. Discovering more information about those objects as necessary.
3. Figuring out information necessary for formatting the display.

(2) and (3) may be dependent on whether using structured or human readable output, but (1) isn't.

Further contributions are:

* moving the ```getattr``` call up to the top level of the action methods and explaining the reason for its existence.
* Renaming ```list_pool```, which was a thoroughly misleading name to ```list_devices```.
* Making calculations of objects to display lazy again.
* Fixing a latent bug lurking in a dependency on a global naming scheme by removing the method containing the bug entirely.
* Using facilities already available in dbus-client-gen, rather than duplicating them, more expensively, in stratis-cli.
* Improving the testing for filesystem list. The testing for blockdev list seemed good enough and I'm leaving better for #303.

Note that there is some gain in algorithmic efficiency by these changes as there are several places where a generator is used instead of a complete list being generated. These changes are more correct as well; the efficiency gains _might_ show up in the presence of very large numbers of blockdevs or filesystems.

I broke this down into small but provably correct edits. This whole change was a bit finicky, so I think the final PR might as well be left with these distinct commits.


mulkieran:
Blocked by #302, mostly because I want to see how this change affects coverage.

mulkieran:
rebased. coverage changes weren't all that interesting, but it was heartening to find the coverage so high.

mulkieran:
I didn't improve the tests of list output, leaving that until later, see #303, but I did a bunch of testing by hand and it all checks out.

PR #305, opened 2019-08-13
Remove unhelpful blackbox tests

mulkieran:
Related: https://github.com/stratis-storage/project/issues/70.

### DIFF
--- a/tests/blackbox/stratis_cli_cert.py
+++ b/tests/blackbox/stratis_cli_cert.py
@@ -21,7 +21,7 @@ import time
 import unittest
 
 from testlib.utils import exec_command, process_exists
-from testlib.stratis import StratisCli, fs_n, p_n
+from testlib.stratis import StratisCli
 
 DISKS = []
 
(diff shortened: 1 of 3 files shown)


mulkieran:
This is up for preliminary review only.

mulkieran:
The last test, for catching KeyboardException should be retained, because it checks real behavior, but moved out of these blackbox tests. It has a sleep time, which must be intended to be long enough for the Python interpreter itself to start up as a result of the call. If it's not long enough, then there will be a test failure, even if our code is correct, like so:

```
$ python3 -c 'while True: pass'
^CTraceback (most recent call last):
  File "<string>", line 1, in <module>
KeyboardInterrupt
```

mulkieran:
rebased, resolved conflicts.

mulkieran:
We should perhaps investigate the ci repo's cleanup scripts, to see if that might not be a better way to cleanup tests. We also should write some quotidian tests for the CLI, doing things that normal users might do via the CLI on a running system.

mulkieran:
rebased.

mulkieran:
Whoops. Test error was a semantic error in a newly added test.

mulkieran:
Ok, should be really fixed this time.

mulkieran:
rebased.

PR #306, opened 2019-08-13
Fix the missing bits in the blackbox tests and enable linting

mulkieran:
The blackbox tests were missing an initialization file, and so pylint wasn't checking them. Now the blackbox tests have an initialization file and they are getting linted.


mulkieran:
I happened to notice this problem while working on #305. I wondered why pylint didn't complain about certain things, and then realized that must be because pylint wasn't being run on the code. Might as well fix all this up before making further changes.

mulkieran:
This is really ready now.

PR #307, opened 2019-08-15
Make blackbox directory not contain a package

mulkieran:


mulkieran:
Related: https://github.com/stratis-storage/project/issues/32 (if we were running these tests as part of the CI my mistake in #306 would have been immediately caught).

Issue #313, opened 2019-08-21
Make computation of auto-generated modules deferred until after module load


mulkieran:
We should probably establish a separate module called ```_generate.py``` in ```_actions```. Then define a method that does what the code block does now. The problem is that the classes that are now just imported from ```_data``` will have to be assigned to some module, somehow.

If the methods are generated when ```_data.py``` is loaded, then an exception raised there and defined by ```stratis_cli``` can not be caught in ```bin/stratis```.

PR #318, opened 2019-08-22
Postpone loading _data.py module until individual actions are executed

mulkieran:
Resolves: #313.

The primary purpose of this PR is to postpone loading the ```_data.py``` module until stratis has begun executing. This allows the exceptions defined in the ```_errors.py``` module to have been loaded previously, so that if an exception is raised while ```_data.py``` is being loaded, the name of the raised exception is available.

Another consequence is that, if the D-Bus timeout is set by an environment variable, under this design, that environment variable would be read only when needed, i.e, when a D-Bus call was about to be made for the first time by the program. I think there is no compelling reason for eager failure on an invalid D-Bus timeout specification, so that seems fine w/ me.

### DIFF
--- a/src/stratis_cli/_actions/__init__.py
+++ b/src/stratis_cli/_actions/__init__.py
@@ -15,7 +15,9 @@
 Package mediating dbus actions.
 """
 
-from ._data import interface_name_to_common_name
+from ._constants import BLOCKDEV_INTERFACE
+from ._constants import FILESYSTEM_INTERFACE
+from ._constants import POOL_INTERFACE
 from ._logical import LogicalActions
 from ._physical import PhysicalActions
 from ._stratis import StratisActions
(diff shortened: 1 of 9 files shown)


mulkieran:
Note that this is solution (1).

Another solution would be to define a method that reads the D-Bus timeout environment variable, and returns the appropriate value for the timeout (or raises an exception). This value is then passed to another function that sets up the module class members (or raises an exception). Both are invoked in the run() method defined in ```_main.py```. This seems a bit more brittle to me than the approach here, because it must operate more by side-effects, i.e., by setting values in the ```_data.py``` module after they are first defined.

A third approach would involve making the classes lazy, so that they construct themselves when they are invoked. But that would be very extra-linguistic, and probably tricky to support.

mulkieran:
rebased. conflicts resolved.

mulkieran:
rebased.

mulkieran:
Ok! I'll tidy this up and get it ready for final review.

mulkieran:
Ok. There's one additional commit "Add an assertion...", but all else is the same.

mulkieran:
rebased.

PR #319, opened 2019-08-22
Make recommendation to use wipefs to free disks more cautious

### LINKED ISSUE #224: Make recommendation to use wipefs to free disks more cautious
It is certainly the case that stratisd might refuse to use a disk because it is a multipath member device. In that case, wiping the device might cause a lot of trouble. So, we can explain that ```wipefs -a``` will remove the signature, but also suggest careful consideration about whether that is the right thing to do.

waltdisgrace:
Resolves #224.

### DIFF
--- a/docs/stratis.txt
+++ b/docs/stratis.txt
@@ -20,8 +20,16 @@ and the filesystems allocated from the pool.
 Stratis creates a *pool* from one or more block devices (*blockdevs*), and
 then enables multiple *filesystems* to be created from the pool.
 
-If the block device is already in use, its signature has to be erased using
-*wipefs -a <blockdev>* before trying to add it to a pool.
+Stratis may refuse to use a block device for any of the following reasons:
+1. The block device could already be in use, in which case its signature must
+be erased using *wipefs -a <blockdev>* before it can be added to a pool.
+2. The block device could be a DM-multipath device, in which case
+*wipefs -a <blockdev>* can also be used, albeit cautiously, as this will
+erase the multipath configuration and cause potentially unwanted side effects.
+
+To determine which case it is, use *multipath -l* to display the current
+multipath configuration gathered from sysfs and the device mapper. If there is
+an entry for the block device in question, then it is a multipath device.
 
 GLOBAL OPTIONS
 --------------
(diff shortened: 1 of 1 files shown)


docs/stratis.txt
```diff
@@ -20,8 +20,16 @@ and the filesystems allocated from the pool.
 Stratis creates a *pool* from one or more block devices (*blockdevs*), and
 then enables multiple *filesystems* to be created from the pool.
 
-If the block device is already in use, its signature has to be erased using
-*wipefs -a <blockdev>* before trying to add it to a pool.
+Stratis may refuse to use a block device for any of the following reasons:
+1. The block device could already be in use, in which case its signature may
```
jbaublitz: For point 1, I'd like to see something even more careful like `The block device could already be in use in another context outside of stratisd, in which case its signature may be erased using *wipefs -a <blockdev>*. To allow use with stratisd, the filesystem must be wiped prior to adding it to a pool, but this should only be done after careful consideration about whether or not the existing data on the device is critical and/or replicated elsewhere.`

Thoughts, @mulkieran?

mulkieran:
We should definitely be cagey. I'll put this back in review, and hope to get a good look at it soon.

docs/stratis.txt
mulkieran: I'ld rather drop it from the description entirely. Its position  makes this text too prominent.

Maybe we can make a separate heading, just below ```ENVIRONMENT VARIABLES```. Call it something very generic like ```NOTES``` for now. Then say something like:

```
If a block device appears to be already in use, stratisd will refuse to claim it. To allow use with
stratisd, any signature on the device must first be erased. Please carefully verify the identity
and availability of the device before taking such a step.
```

I'ld rather leave multipath specifically out of the discussion entirely, basically because there are so many other tricky possibilities besides multipath to worry about.

docs/stratis.txt
```diff
@@ -85,6 +82,13 @@ STRATIS_DBUS_TIMEOUT::
          which represents the timeout length in milliseconds
          2. -1, which represents the libdbus default timeout
 
+NOTES
+-----
+If a block device appears to be already in use, stratisd will refuse to
+claim it. To allow use with stratisd, any signature on the device must
+first be erased. Please carefully verify the identityand availability of
```
jbaublitz: I think you meant `identity and`.

mulkieran:
@waltdisgrace Please fix up typo, I hope to ge tthis merged today.

PR #297, opened 2019-07-31
Changes to CLI to support changes in stratisd engine and DBus API

jbaublitz:
This is related to stratis-storage/project#51 and stratis-storage/stratisd#1597. This is simply the CLI-side work.

PR #321, opened 2019-08-22
Convert no-op engine results into failures

jbaublitz:
Supersedes #297 

I moved the branch to my fork and am kicking off CI again as that seems to be hanging even on new commits.

@mulkieran I'll address the exception issue you mentioned in the last PR in this new PR. The other [issue you mentioned](https://github.com/stratis-storage/stratis-cli/pull/297#discussion_r310702276) about the check I was doing for idempotence has already been resolved.

### DIFF
--- a/src/stratis_cli/_actions/_logical.py
+++ b/src/stratis_cli/_actions/_logical.py
@@ -19,8 +19,8 @@ from justbytes import Range
 from dateutil import parser as date_parser
 
 from .._errors import StratisCliEngineError
-from .._errors import StratisNoChangeError
-from .._errors import StratisPartialChangeError
+from .._errors import StratisCliNoChangeError
+from .._errors import StratisCliPartialChangeError
 
 from .._stratisd_constants import StratisdErrors
 
@@ -67,7 +67,7 @@ class LogicalActions:
         already_names = requested_names.intersection(names)
 
         if already_names != frozenset():
-            raise StratisPartialChangeError(
+            raise StratisCliPartialChangeError(
                 "create",
                 list(requested_names.difference(already_names)),
                 list(already_names),
@@ -174,7 +174,7 @@ class LogicalActions:
         already_removed = requested_names.difference(frozenset(pool_filesystems.keys()))
 
         if already_removed != frozenset():
-            raise StratisPartialChangeError(
+            raise StratisCliPartialChangeError(
                 "destroy",
                 requested_names.difference(already_removed),
                 list(already_removed),
@@ -227,7 +227,7 @@ class LogicalActions:
 
         if not changed:
             resource = namespace.snapshot_name
-            raise StratisNoChangeError("snapshot", resource)
+            raise StratisCliNoChangeError("snapshot", resource)
 
     @staticmethod
     def rename_fs(namespace):
@@ -261,4 +261,4 @@ class LogicalActions:
 
         if not changed:
             resource = namespace.new_name
-            raise StratisNoChangeError("rename", resource)
+            raise StratisCliNoChangeError("rename", resource)
--- a/src/stratis_cli/_actions/_logical.py
+++ b/src/stratis_cli/_actions/_logical.py
@@ -222,8 +222,7 @@ class LogicalActions:
             raise StratisCliEngineError(rc, message)
 
         if not changed:
-            resource = namespace.snapshot_name
-            raise StratisCliNoChangeError("snapshot", resource)
+            raise StratisCliNoChangeError("snapshot", namespace.snapshot_name)
 
     @staticmethod
     def rename_fs(namespace):
@@ -256,5 +255,4 @@ class LogicalActions:
             raise StratisCliEngineError(rc, message)
 
         if not changed:
-            resource = namespace.new_name
-            raise StratisCliNoChangeError("rename", resource)
+            raise StratisCliNoChangeError("rename", namespace.new_name)
--- a/src/stratis_cli/_actions/_logical.py
+++ b/src/stratis_cli/_actions/_logical.py
@@ -68,9 +68,7 @@ class LogicalActions:
 
         if already_names != frozenset():
             raise StratisCliPartialChangeError(
-                "create",
-                list(requested_names.difference(already_names)),
-                list(already_names),
+                "create", requested_names.difference(already_names), already_names
             )
 
         (_, rc, message) = Pool.Methods.CreateFilesystems(
@@ -175,9 +173,7 @@ class LogicalActions:
 
         if already_removed != frozenset():
             raise StratisCliPartialChangeError(
-                "destroy",
-                requested_names.difference(already_removed),
-                list(already_removed),
+                "destroy", requested_names.difference(already_removed), already_removed
             )
 
         fs_object_paths = [
--- a/src/stratis_cli/_actions/_logical.py
+++ b/src/stratis_cli/_actions/_logical.py
@@ -19,6 +19,8 @@ from justbytes import Range
 from dateutil import parser as date_parser
 
 from .._errors import StratisCliEngineError
+from .._errors import StratisNoChangeError
+from .._errors import StratisPartialChangeError
 
 from .._stratisd_constants import StratisdErrors
 
@@ -39,8 +41,11 @@ class LogicalActions:
 
         :raises StratisCliEngineError:
         """
+        # pylint: disable=too-many-locals
+        from ._data import MOFilesystem
         from ._data import ObjectManager
         from ._data import Pool
+        from ._data import filesystems
         from ._data import pools
 
         proxy = get_object(TOP_OBJECT)
@@ -50,6 +55,16 @@ class LogicalActions:
             .require_unique_match(True)
             .search(managed_objects)
         )
+
+        fs_names = {
+            MOFilesystem(info).Name(): path
+            for (path, info) in filesystems().search(managed_objects)
+        }
+        unchanged = [name for name in namespace.fs_name if name in fs_names]
+
+        if unchanged == []:
+            changed = [name for name in namespace.fs_name if name not in fs_names]
+            raise StratisPartialChangeError("create", changed, unchanged)
 
         (_, rc, message) = Pool.Methods.CreateFilesystems(
             get_object(pool_object_path), {"specs": namespace.fs_name}
@@ -125,6 +140,8 @@ class LogicalActions:
 
         :raises StratisCliEngineError:
         """
+        # pylint: disable=too-many-locals
+        from ._data import MOFilesystem
         from ._data import ObjectManager
         from ._data import Pool
         from ._data import filesystems
@@ -145,6 +162,16 @@ class LogicalActions:
                 props={"Name": name, "Pool": pool_object_path}
             ).search(managed_objects)
         ]
+
+        fs_names = [
+            MOFilesystem(info).Name()
+            for (_, info) in filesystems().search(managed_objects)
+        ]
+        unchanged = [name for name in namespace.fs_name if name not in fs_names]
+
+        if unchanged == []:
+            changed = [name for name in namespace.fs_name if name in fs_names]
+            raise StratisPartialChangeError("destroy", changed, unchanged)
 
         (_, rc, message) = Pool.Methods.DestroyFilesystems(
             get_object(pool_object_path), {"filesystems": fs_object_paths}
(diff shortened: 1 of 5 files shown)


jbaublitz:
ok to test

mulkieran:
add to whitelist

jbaublitz:
@bgurney-rh These are expected test failures. I've edited them to expect the correct number of return arguments from the new DBus interface, but we're still using the old version of stratisd in tests.

src/stratis_cli/_actions/_logical.py
```diff
@@ -54,10 +54,15 @@ def create_volumes(namespace):
             .search(managed_objects)
         )
 
-        (_, rc, message) = Pool.Methods.CreateFilesystems(
+        ((changed, _), rc, message) = Pool.Methods.CreateFilesystems(
             get_object(pool_object_path), {"specs": namespace.fs_name}
         )
 
+        if not changed:
```
mulkieran: I _think_ the order of this code block and the next one should be reversed, i.e., check the value of ```rc``` first, and _then_ check ```changed```.

jbaublitz:
The exception is up for review and you're absolutely right, the return code check becomes useless if the changed value is checked first as it's always `false` in the `default_return` in the engine. Thanks for catching that!

src/stratis_cli/_errors.py
```diff
@@ -30,6 +30,25 @@ class StratisCliRuntimeError(StratisCliError):
     """
 
 
+class StratisNoActionError(StratisCliError):
```
mulkieran: Could you call this ```NoChange``` rather than ```NoAction```? ```Action``` is sort of already taken, see below.

src/stratis_cli/_errors.py
```diff
@@ -30,6 +30,83 @@ class StratisCliRuntimeError(StratisCliError):
...
+    Raised if a block device is added as both a blockdev and a cachedev
+    """
+
+    def __init__(self, blockdevs, added_as="cache"):
+        """ Initializer.
+
+            :param list of str blockdevs: the blockdevs that would be added in both tiers
+            :param str added_as: whether the action causing the error on the cache or data tier
+        """
+        super().__init__()
```
mulkieran: Do not call the initializer...use the ```#pylint: disable``` directive instead. Call the initializer and the auto-generated ```__repr__``` is incorrect. I'm not sure why this is, but this is definitely the case and appears to be the original motivation for the way the other exceptions were set up.

src/stratis_cli/_actions/_top.py
```diff
@@ -153,9 +172,32 @@ def add_data_devices(namespace):
...
 
+        cache = [
+            MODev(info).Devnode()
+            for (_, info) in devs(props={"Tier": BlockDevTiers.Cache}).search(
+                managed_objects
+            )
+        ]
+        already_cache = [dev for dev in namespace.blockdevs if dev in cache]
+        if already_cache != []:
+            raise StratisInUseError(already_cache, "data")
```
mulkieran: Please make up an enum constant for data and cache and put it in ```_action/_constants.py```. The constant will ultimately suit better than a string.

src/stratis_cli/_actions/_top.py
```diff
@@ -153,9 +172,32 @@ def add_data_devices(namespace):
             .search(managed_objects)
         )
 
+        cache = [
```
mulkieran: I think it makes sense to use ```frozenset``` to make it obvious that these are a bunch of set operations.

```
blockdevs = frozenset(namespace.blockdevs)
cache = frozenset(...)
already_cache =  blockdevs intersection cache

unchanged = blockdevs intersection data
# etc.
```

Also, I think ```unchanged``` would be clearer as ```already_data``` to go with ```already_cache```.

src/stratis_cli/_errors.py
jbaublitz: Resolved.

src/stratis_cli/_actions/_top.py
jbaublitz: There is an existing enum in `_stratisd_constants.py` so I've switched over to using that.

src/stratis_cli/_actions/_top.py
jbaublitz: I've pushed a bug fix and the name change so far. Still working on the rest.

jbaublitz: This is now resolved. I've switched to `frozenset`, convert to lists of strings for output, and use the appropriate set operations.

src/stratis_cli/_actions/_top.py
```diff
@@ -172,27 +172,31 @@ def add_data_devices(namespace):
...
             for (_, info) in devs(props={"Tier": BlockDevTiers.Cache}).search(
                 managed_objects
             )
-        ]
-        already_cache = [dev for dev in namespace.blockdevs if dev in cache]
-        if already_cache != []:
-            raise StratisInUseError(already_cache, BlockDevTiers.Data)
+        )
+        already_cache = blockdevs.intersection(cache)
+        if list(already_cache) != []:
```
mulkieran: Don't convert it to a list...just compare with empty frozenset.

src/stratis_cli/_actions/_top.py
```diff
@@ -172,27 +172,31 @@ def add_data_devices(namespace):
...
             )
-        ]
-        already_data = [dev for dev in namespace.blockdevs if dev in data]
+        )
+        already_data = blockdevs.intersection(data)
 
-        if already_data != []:
-            changed = [dev for dev in namespace.blockdevs if dev not in already_data]
-            raise StratisPartialChangeError("add-data", changed, already_data)
+        if list(already_data) != []:
```
mulkieran: compare with empty frozenset here, also.

src/stratis_cli/_actions/_top.py
```diff
@@ -172,27 +172,31 @@ def add_data_devices(namespace):
...
+            changed = blockdevs.difference(already_data)
```
mulkieran: ```changed``` -> ```not_data``` ?

src/stratis_cli/_errors.py
```diff
@@ -30,6 +32,87 @@ class StratisCliRuntimeError(StratisCliError):
...
+class StratisInUseError(StratisCliError):
+    """
+    Raised if a block device is added as both a blockdev and a cachedev
+    """
+
+    def __init__(self, blockdevs, added_as="cache"):
+        """ Initializer.
+
+            :param list of str blockdevs: the blockdevs that would be added in both tiers
+            :param str added_as: whether the action causing the error on the cache or data tier
```
mulkieran: Please update the docs for ```added_as``` to correspond w/ the recent changes. Also, please make ```added_as``` a mandatory argument. The code for ```__str__``` will be shorter, and since this exception isn't part of any public API, we don't need to worry about future extensibility.

src/stratis_cli/_errors.py
```diff
@@ -30,6 +32,87 @@ class StratisCliRuntimeError(StratisCliError):
...
+
+class StratisInUseError(StratisCliError):
+    """
+    Raised if a block device is added as both a blockdev and a cachedev
+    """
+
+    def __init__(self, blockdevs, added_as="cache"):
+        """ Initializer.
+
+            :param list of str blockdevs: the blockdevs that would be added in both tiers
```
mulkieran: I have always believed that if you want a complicated type you have to do it separately, like:

```
:param blockdevs: ...
:type blockdevs: list of str
```
if this isn't true anymore, that would be nice, too.

jbaublitz: Uh oh! I lifted that syntax from somewhere else in the code so the doc string issue may be a little more complicated than originally thought if that's not valid syntax. I'll dig into it a bit more.

src/stratis_cli/_actions/_top.py
jbaublitz: I'm not sure that would be a clear variable name to me personally. I could see something like `new_data` for sure if that's acceptable?

src/stratis_cli/_errors.py
jbaublitz: The documentation for type hints does seem to indicate that type can go [after `:param` and before `name_of_parameter:`.](https://github.com/agronholm/sphinx-autodoc-typehints/blob/master/README.rst#sphinx-autodoc-typehints)

jbaublitz: Resolved.

src/stratis_cli/_errors.py
jbaublitz: Resolved.

src/stratis_cli/_actions/_top.py
jbaublitz: Resolved.

src/stratis_cli/_actions/_top.py
jbaublitz: Resolved.

src/stratis_cli/_actions/_top.py
jbaublitz: Resolved.

src/stratis_cli/_errors.py
mulkieran: I went to the source but I don't think that usage is available. Here: https://www.sphinx-doc.org/en/1.8/usage/restructuredtext/domains.html?highlight=%3Atype#the-python-domain. It seems like:

```
:param <name>: <description>
:type <name>: <the type>
```

is the standard and

```
:param <type> <name>: <description>
```
is a shorthand when the type is simple. Quoting: "It is also possible to combine parameter type and description, if the type is a single word, like this..." from the docs. So, I think that clinches it.

jbaublitz:
Rebased and ready for review as discussed.

mulkieran:
@jbaublitz I found out that some semantic changes did seem to be indicated after all. I've put them up here: https://github.com/stratis-storage/stratis-cli/pull/342. Please review them carefully before accepting them, as they do change behaviour.

jbaublitz:
Rebased in preparation for incorporating your review.

jbaublitz:
@mulkieran Updated with your edits.

mulkieran:
@jbaublitz Please see new commits in https://github.com/stratis-storage/stratis-cli/pull/342. I cherry-picked them from a larger set that I have stashed; but these seem to make a coherent whole.

jbaublitz:
@mulkieran This is now rebased and your commits have been squashed into two revision PRs corresponding to the two targeted areas that you've changed. I ran some tests manually, reviewed the code, and overall think this looks good so I've pushed it to my PR.

mulkieran:
@jbaublitz: Please take a look at changes in #342. I believe that all are refactorings or rewordings, except the very last, which adds one additional exception class.

jbaublitz:
@mulkieran All of this looks good, but I am a little confused on why we're duplicating the functionality of the idempotence in the engine by reimplementing a check upfront in the CLI in `_check_same_tier()`. My guess is that this because of the addition of the `IncoherenceError` which, due to `GetManagedObjects` being involved, could make it unclear if stratis-cli is experiencing a race condition or whether the device being added already existed. If this is the rationale, I think I'm ready to pull all of the changes over to this PR.

mulkieran:
@jbaublitz: ```_check_same_tier``` is not intended to introduce any new behavior. ```_check_opposite_tier``` raises ```InUseError``` if it finds any blockdev already used in the opposite tier, ```_check_same_tier``` raises ```PartialChange``` if it finds any blockdev already used in the chosen tier.

```IncoherenceError``` covers a different, and somewhat unusual case.

There's a symmetry in the whole change: methods can either raise ```NoChangeError```, in the simple case where no pre-checking is required before invoking the D-Bus method, or they can raise both of ```PartialChange``` and ```IncoherenceError``` in the case where there is a potential action on a set of some sort.

**Comment by jbaublitz** [comment]
Whoops! I seem to have gotten confused. Incorporating these now.

**Comment by mulkieran** [comment]
Another unnecessary comprehension seems to be sitting around.

**Comment by jbaublitz** [comment]
This one is actually not unnecessary. It is retrieving values from a generator and collecting them into a list. Disabling this lint now.

File: src/stratis_cli/_error_reporting.py [hunk 11]
```diff
@@ -205,6 +205,7 @@ def handle_error(err):
     :param Exception err: an exception
     """
 
+    # pylint: disable=unnecessary-comprehension
```
mulkieran: You can just use ```list(get_errors(err))```

jbaublitz: Oops! That was a rebase failure on my part. It was already resolved in master. I've rebased and removed that commit.

PR #328, opened 2019-08-28
Get environment variable and pass into make_class function

waltdisgrace:


src/stratis_cli/_actions/_data.py
```diff
@@ -207,29 +209,54 @@
 DBUS_TIMEOUT_SECONDS = 120
 
 
+def get_timeout():
+    """Get timeout environment variable and return it as an integer"""
```
mulkieran: Please format the method doc string as is done elsewhere. We use sphinx markup to generate docs, see; this is why you see a lot of ```:param``` and ```:return``` strings in the docs. Also, ```:raises```.

mulkieran:
@waltdisgrace Notice that at this point, your code is not only behind master but has conflicts. As soon as is convenient, you should go ahead, rebase on current master, and resolve your conflicts in the best way, i.e., the one that makes semantic sense. It is reasonable to first review your own PR and also review the PR that induced the conflict _before_ you begin the rebase process, as this might help you to understand how best to resolve it. _Don't_ use the "Resolve Conflicts" button in GitHub to resolve the conflicts, but it's not wrong to press the button and look at the display that GitHub gives you, so that you can understand where the conflict is in order to prepare for your rebase step.

waltdisgrace:
rebased.

mulkieran:
Ok! Now, go ahead and generate a stack trace with a bad value for the environment variable and paste it up here.

waltdisgrace:
```
[gchin@to-be-determined stratis-cli]$ PYTHONPATH=./src STRATIS_DBUS_TIMEOUT="-2" ./bin/stratis pool list
  File "./bin/stratis", line 35, in <module>
    main()
  File "./bin/stratis", line 31, in main
    return run()(sys.argv[1:])
  File "/home/gchin/stratis-storage/stratis-cli/src/stratis_cli/_main.py", line 52, in the_func
    raise err
  File "/home/gchin/stratis-storage/stratis-cli/src/stratis_cli/_main.py", line 43, in the_func
    result.func(result)
  File "/home/gchin/stratis-storage/stratis-cli/src/stratis_cli/_actions/_top.py", line 65, in list_pools
    from ._data import MOPool
  File "/home/gchin/stratis-storage/stratis-cli/src/stratis_cli/_actions/_data.py", line 252, in <module>
    timeout = get_timeout(timeout_string)
  File "/home/gchin/stratis-storage/stratis-cli/src/stratis_cli/_actions/_data.py", line 233, in get_timeout
    "The timeout value you provided exceeds the smallest acceptable value, -1."
stratis_cli._errors.StratisCliEnvironmentError: The timeout value you provided exceeds the smallest acceptable value, -1.

**Comment by waltdisgrace** [comment]
```
[gchin@to-be-determined stratis-cli]$ PYTHONPATH=./src STRATIS_DBUS_TIMEOUT="-2.0" ./bin/stratis pool list
  File "/home/gchin/stratis-storage/stratis-cli/src/stratis_cli/_actions/_data.py", line 223, in get_timeout
    timeout_int = int(input_string)
ValueError: invalid literal for int() with base 10: '-2.0'

During handling of the above exception, another exception occurred:

  File "./bin/stratis", line 35, in <module>
    main()
  File "./bin/stratis", line 31, in main
    return run()(sys.argv[1:])
  File "/home/gchin/stratis-storage/stratis-cli/src/stratis_cli/_main.py", line 52, in the_func
    raise err
  File "/home/gchin/stratis-storage/stratis-cli/src/stratis_cli/_main.py", line 43, in the_func
    result.func(result)
  File "/home/gchin/stratis-storage/stratis-cli/src/stratis_cli/_actions/_top.py", line 65, in list_pools
    from ._data import MOPool
  File "/home/gchin/stratis-storage/stratis-cli/src/stratis_cli/_actions/_data.py", line 252, in <module>
    timeout = get_timeout(timeout_string)
  File "/home/gchin/stratis-storage/stratis-cli/src/stratis_cli/_actions/_data.py", line 227, in get_timeout
    "The timeout value you provided is not an integer."
stratis_cli._errors.StratisCliEnvironmentError: The timeout value you provided is not an integer.

mulkieran:
Excellent. Notice that for both error chains that you recorded the stack trace exits at line 35 in ```./bin/stratis```. Use an ```except``` statement to catch the exception in the same way that the ```KeyboardException``` is caught there, and choose some error message to exit with. Don't worry too much about whether it's the best message.

tests/whitebox/integration/test_timeout.py
```diff
@@ -0,0 +1,91 @@
...
+class TimeoutTestCase(unittest.TestCase):
+    """
+    Test various timeout inputs.
+    """
+
+    def testTimeoutTooLarge(self):
+        """
+        Should fail because the timeout value is too large.
+        """
+        test_value = "2147483648"
```
mulkieran: The whole body of the test can be condensed into:

```
with self.assertRaises(StratisCliEnvironmentError):
    get_timeout("2147483648")
```

which is more correct. The problem with the way your test is setup is that if an exception that is not the expected one is raised, the test framework sees a test error, not a test failure. Go ahead and update all the other similar tests, similarly.

bin/stratis
```diff
@@ -36,3 +37,7 @@ try:
 
 except KeyboardInterrupt:
     sys.exit("Received KeyboardInterrupt")
+
+except StratisCliEnvironmentError:
+    sys.exit("Received StratisCliEnvironmentError")
+
```
mulkieran: extra lines!

src/stratis_cli/_actions/_data.py
```diff
@@ -207,29 +209,68 @@
...
+        )
+
+    # Ensure the integer is not too large
+    if timeout_int > (1 << 31) - 1:
+        raise StratisCliEnvironmentError(
+            "The timeout value you provided exceeds the largest acceptable value, 2147483647."
+        )
+
+    # Convert from milliseconds to seconds
+    timeout_float = timeout_int / 1000
```
mulkieran: You can return a whole expression, and it's a bit more succinct:

```
return timeout_int / 1000
```

is a bit better than these two lines.

src/stratis_cli/_actions/_data.py
```diff
@@ -207,29 +209,68 @@
...
+
+    return timeout_float
+
+
 try:
+
+    timeout = environ.get("STRATIS_DBUS_TIMEOUT", DBUS_TIMEOUT_SECONDS * 1000)
```
mulkieran: It would be nice to collapse these two lines into one, also.

tests/whitebox/integration/test_timeout.py
mulkieran: @waltdisgrace It looks like you're not using ```assertRaises``` where you should be.

waltdisgrace:
rebased.

tests/whitebox/integration/test_timeout.py
```diff
@@ -0,0 +1,101 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Test 'timeout'.
+"""
+
+import unittest
+
+from stratis_cli._errors import StratisCliEnvironmentError
+
```
mulkieran: Eliminate this line, since both imports are from stratis_cli.

tests/whitebox/integration/test_timeout.py
```diff
@@ -0,0 +1,101 @@
+# Copyright 2016 Red Hat, Inc.
```
mulkieran: Move this entire file to a new directory "tests/whitebox/unit/test_timeout.py". In order for the tests to run, you'll also need to add an empty ```__init__.py``` file in the "tests/whitebox/unittest" directory.

src/stratis_cli/_actions/_data.py
```diff
@@ -207,29 +209,71 @@
...
+        )
+
+    # Ensure the integer is not too small
+    if timeout_int < -1:
+        raise StratisCliEnvironmentError(
+            "The timeout value you provided is smaller than the smallest acceptable value, -1."
+        )
+
+    # Ensure the integer is not too large
+    if timeout_int > (1 << 30) - 1:
```
mulkieran: Make this value a constant, and call it something like ```MAXIMUM_DBUS_TIMEOUT_MS```. You can put the definition next to the ```DBUS_TIMEOUT_SECONDS``` definition. You should use the constant value in the error  message, also.

bin/stratis
```diff
@@ -36,3 +37,6 @@ try:
 
 except KeyboardInterrupt:
     sys.exit("Received KeyboardInterrupt")
+
+except StratisCliEnvironmentError:
+    sys.exit("Received StratisCliEnvironmentError")
```
mulkieran: You'll want some more helpful error here, like:

```Value of STRATIS_DBUS_TIMEOUT invalid: %s" % str(<the specific message associated with the error caught>)```

File: src/stratis_cli/_actions/_data.py [hunk 10]
```diff
@@ -27,6 +28,7 @@
 from dbus_python_client_gen import DPClientGenerationError
 
 from .._errors import StratisCliGenerationError
+from .._errors import StratisCliEnvironmentError
```
mulkieran: alphabetize! E before G. Hopefully soon we'll be handing this detail off to a program.

docs/stratis.txt
```diff
@@ -75,6 +75,17 @@ daemon redundancy::
 daemon version::
        Show the Stratis service's version.
 
+ENVIRONMENT VARIABLES
+---------------------
+STRATIS_DBUS_TIMEOUT::
+	 Sets a timeout for any Stratis D-Bus call. If this environment variable is
+	 not set, a default value of 120 seconds is used for the timeout.
+	 Accepted STRATIS_DBUS_TIMEOUT environment variable values are:
+         1. an integer between 0 (inclusive) and INT31_MAX (inclusive),
```
mulkieran: Someone might object that INT31_MAX is not a widely understood name, we should probably come up with some different way of saying the same thing.

mulkieran:
@waltdisgrace Once you move your test from ```integration``` subdirectory to the ```unittest``` subdirectory you'll have to do a few other things as well. You should make a new Makefile target, similar to the ones for ```dbus-tests``` and ```keyboard-interrupte-test```. You should add making the unittests to the coverage test, just like the ```keyboard-interrupt-test```.

You should add your new Makefile target to the Travis configuration file as well. But you'll want to make sure that the other steps are working first before you edit the Travis file.

You won't need to update anything in the ci repo, because your unit tests will do just fine just being run on Travis.

src/stratis_cli/_actions/_data.py
```diff
@@ -205,31 +207,76 @@
...
+    # Ensure the integer is not too small
+    if timeout_int < -1:
+        raise StratisCliEnvironmentError(
+            "The timeout value you provided is smaller than the smallest acceptable value, -1."
+        )
+
+    # Ensure the integer is not too large
+    if timeout_int > MAXIMUM_DBUS_TIMEOUT_MS:
+        raise StratisCliEnvironmentError(
+            "The timeout value you provided exceeds the largest acceptable value, "
```
mulkieran: Use string formatting here:

```The timeout value...largest acceptable value, %s." % MAXIMUM_DBUS_TIMEOUT_MS```

it'll apply the string formatting automatically, since you're using ```%s```.

File: bin/stratis [hunk 13]
```diff
@@ -22,6 +22,7 @@ try:
     import sys
 
     from stratis_cli import run
+    from stratis_cli._errors import StratisCliEnvironmentError
```
mulkieran: Export ```StratisCliEnvironmentError``` in ```stratis_cli/__init__.py```, and import it directly from stratis_cli here. You should do the same in ```test_timeout.py```.

bin/stratis
```diff
@@ -36,3 +37,6 @@ try:
 
 except KeyboardInterrupt:
     sys.exit("Received KeyboardInterrupt")
+
+except StratisCliEnvironmentError as err:
+    sys.exit("Value of STRATIS_DBUS_TIMEOUT invalid: %s" % str(err))
```
mulkieran: Don't use ```str``` here.  Using ```%s``` automatically invokes ```str```.

docs/stratis.txt
```diff
@@ -75,6 +75,17 @@ daemon redundancy::
...
 
+ENVIRONMENT VARIABLES
+---------------------
+STRATIS_DBUS_TIMEOUT::
+	 Sets a timeout for any Stratis D-Bus call. If this environment variable is
+	 not set, a default value of 120 seconds is used for the timeout.
+	 Accepted STRATIS_DBUS_TIMEOUT environment variable values are:
+         1. an integer between 0 (inclusive) and 1073741823 (inclusive),
+         which represents the timeout length in milliseconds
+         2. -1, which represents a default that was
```
mulkieran: Instead of "a default that was supplied by libdbus", use "the libdbus default timeout".

mulkieran:
Also, I've added a Travis task for running the unittests, and it is now working: https://github.com/stratis-storage/stratis-cli/pull/340/commits/ac088fc7d1ee1572d1c265563a25b24a4dc66bc8. It's in PR: https://github.com/stratis-storage/stratis-cli/pull/340.

You should be able to add this commit onto the end of your PR. The means I would use would be to checkout the PR that contains the commit and cherry pick the commit onto my branch. I'm sure there are others.

PR #329, opened 2019-08-28
Edit Makefile targets to use isort and apply formatting to Python files

jbaublitz:
This is a potential partial resolution for stratis-storage/project#40. It adds `isort` invocations into `make fmt` and `make fmt-travis` using single line sorted imports. This pull request also includes application of `make fmt` using `isort`, and I am open to feedback on if this is indeed what we want.


mulkieran:
We'll push this back to October, as isort is changing its ways w/ different versions of Python.

jbaublitz:
Interestingly, this seems to work now. @mulkieran Is it worth taking another look at this?

bgurney-rh:
Error from travis:
```
ERROR: /home/travis/build/stratis-storage/stratis-cli/tests/blackbox/stratisd_cert.py Imports are incorrectly sorted.
```

Should the "import dbus" line in stratisd_cert.py be moved between "import argparse" and "import sys"?

jbaublitz:
I may have found the issue. Will report back after looking into it more.

PR #333, opened 2019-08-30
Fix a test

mulkieran:


tests/whitebox/integration/test_stratis.py
```diff
@@ -92,11 +92,20 @@ class ErrorHandlingTestCase(SimTestCase):
...
+        exit.
         """
-        command_line = ["pool", "list", "not_existing"]
+        command_line = ["filesystem", "list", "not_existing"]
+
+        # if exceptions are propagated then a Stratis error is caught
+        with self.assertRaises(StratisCliActionError):
+            RUNNER(["--propagate"] + command_line)
+
+        # If instead the exception chain is handed of to handle_error,
```
bgurney-rh: "If instead the exception chain is handed off to handle_error,"?

mulkieran: ```handle_error``` is the method that tries to understand the exception chain if the exception is not just propagated.

jbaublitz: I think it's more a question of if it's off or of. I was wondering if that's a typo too.

mulkieran: Oh! I'll fix that typo.

mulkieran:
typo fixed.

PR #337, opened 2019-09-09
Cleanup blackbox tests

mulkieran:
This PR tidies up the existing blackbox testing infrastructure and inserts two example tests one for stratisd and one for stratis-cli.

It establishes an exec_test_command method to be used to run tests. This method contains no internal assertions, it is expected that the test that calls it will make the assertions about expected behavior and they'll use the appropriate unittest methods to do so.

It simplifies exec_command which is now used only for infrastructure actions, so that it always expects success.

It also fixes up a bunch of pylint errors, taking advantage of the point where the amount of code is a minimum to remove some blanket disables from check.py. It may be necessary to add some of these blanket disables back in again as the test suite is expanded. However, they were all previously grandfathered in from when these files weren't being pylinted, so it may prove not necessary, also.

There are a bunch of StratisCli methods remaining in the stratis module. These are used in the test infrastructure for setting up and cleaning up actions. We must make these methods independent of the daemon and the CLI, especially for the daemon tests, but I'm leaving them alone for now.

It also does not take steps to do any abstractions for either the busctl or the stratis commands although they are likely to be repetitious. This abstraction is also deferred until a later PR. The current tests in both files just constitute reasonable examples to build on.

Related:
https://github.com/stratis-storage/project/issues/45
https://github.com/stratis-storage/project/issues/79
https://github.com/stratis-storage/project/issues/80

### DIFF
--- a/check.py
+++ b/check.py
@@ -25,8 +25,6 @@ arg_map = {
     "tests/blackbox/testlib": [
         "--reports=no",
         "--disable=I",
-        "--disable=duplicate-code",
-        "--disable=invalid-name",
         "--msg-template='{path}:{line}: [{msg_id}({symbol}), {obj}] {msg}'",
     ],
     "tests/whitebox": [
--- a/check.py
+++ b/check.py
@@ -15,13 +15,11 @@ arg_map = {
     "tests/blackbox/stratisd_cert.py": [
         "--reports=no",
         "--disable=I",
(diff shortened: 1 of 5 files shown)


mulkieran:
rebased.

mulkieran:
Whoops. I noticed a bug.

mulkieran:
squashed.

PR #342, opened 2019-09-11
Jbaublitz issue project 51

mulkieran:


src/stratis_cli/_actions/_logical.py
```diff
@@ -138,12 +162,26 @@ def destroy_volumes(namespace):
...
+        already_removed = requested_names.difference(frozenset(pool_filesystems.keys()))
+
+        if already_removed != frozenset():
+            raise StratisPartialChangeError(
+                "destroy",
+                requested_names.difference(already_removed),
+                list(already_removed),
+            )
+
         fs_object_paths = [
```
jbaublitz: This make sense so we don't do unnecessary work if we already know it will fail.

jbaublitz:
It seems like it'll be hard to inline comments so I'll post my reactions here based on the changes I see:
* The changes to `create_volumes()` seem to make sense. These seem to mainly be a switch to frozenset and the order of the operations have stayed the same.
* The changes to `destroy_volumes()` seem to have one change in order of operations which makes sense. It looks like the search is done more broadly on all filesystems for the pool and then the check is done if any of the provided removals have already been removed. After that, the object paths are calculated from the original search. This makes sense to me as we don't need to calculate object paths if we already know we'll fail to execute the command before that point.

Overall these look good to me. I'll rebase my branch and then incorporate them into my PR. Would you like to squash those before I do that?

jbaublitz:
I've updated my CLI PR with your squashed edits.

mulkieran:
All addressed in #358.

PR #344, opened 2019-09-11
Remove dependency on stratis cli in blackbox tests set up and tear down

jbaublitz:
Closes stratis-storage/project#81

The branch naming for my source branch would've confused me so I'm recreating #343 with the proper source branch.

### DIFF
--- a/tests/blackbox/testlib/dbus.py
+++ b/tests/blackbox/testlib/dbus.py
@@ -51,13 +51,9 @@ class StratisDbus:
         Query the pools
         :return: A list of pool names.
         """
-        objects = StratisDbus._get_managed_objects().items()
-        if objects is None:
-            objects = {}
-
         pool_objects = [
             obj_data[StratisDbus._POOL_IFACE]
-            for _, obj_data in objects
+            for _, obj_data in StratisDbus._get_managed_objects().items()
             if StratisDbus._POOL_IFACE in obj_data
         ]
 
@@ -70,29 +66,25 @@ class StratisDbus:
         :param name: Name of pool to destroy
         :return: None
         """
-        objects = StratisDbus._get_managed_objects().items()
-        if objects is None:
-            objects = {}
-
         pool_objects = {
             path: obj_data[StratisDbus._POOL_IFACE]
-            for path, obj_data in objects
+            for path, obj_data in StratisDbus._get_managed_objects().items()
             if StratisDbus._POOL_IFACE in obj_data
         }
 
-        pool_object_path = [
+        pool_object_paths = [
             path
             for path, pool_obj in pool_objects.items()
             if pool_obj["Name"] == pool_name
-        ].pop()
-        if pool_object_path is None:
+        ]
+        if pool_object_paths == []:
             return None
 
         iface = dbus.Interface(
             StratisDbus._BUS.get_object(StratisDbus._BUS_NAME, StratisDbus._TOP_OBJECT),
             StratisDbus._MNGR_IFACE,
         )
-        iface.DestroyPool(pool_object_path)
+        iface.DestroyPool(pool_object_paths[0])
 
         return None
 
@@ -105,8 +97,6 @@ class StratisDbus:
                                 CREATED_TS]
         """
         objects = StratisDbus._get_managed_objects().items()
-        if objects is None:
-            objects = {}
 
         fs_objects = [
             obj_data[StratisDbus._FS_IFACE]
@@ -120,18 +110,10 @@ class StratisDbus:
             if StratisDbus._POOL_IFACE in obj_data
         }
 
-        result = {}
-        for fs_object in fs_objects:
-            created = fs_object["Created"]
-            values = {
-                "POOL_NAME": pool_path_to_name[fs_object["Pool"]],
-                "USED_SIZE": fs_object["Used"],
-                "UUID": fs_object["Uuid"],
-                "SYM_LINK": fs_object["Devnode"],
-                "CREATED": created,
-                "CREATED_TS": created,
-            }
-            result[fs_object["Name"]] = values
+        result = {
+            fs_object["Name"]: pool_path_to_name[fs_object["Pool"]]
+            for fs_object in fs_objects
+        }
 
         return result
 
@@ -144,8 +126,6 @@ class StratisDbus:
         :return: None
         """
         objects = StratisDbus._get_managed_objects().items()
-        if objects is None:
-            objects = {}
 
         pool_objects = {
             path: obj_data[StratisDbus._POOL_IFACE]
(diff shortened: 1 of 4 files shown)


jbaublitz:
The dependency should now be removed in stratisd blackbox tests. I'm going to try to move forward with directly interacting with device mapper for the `clean_up()` function.

jbaublitz:
Putting this in preliminary review for now.

tests/blackbox/testlib/dbus.py
```diff
@@ -0,0 +1,207 @@
...
+        pool_path_to_name = {
+            obj: obj_data[StratisDbus._POOL_IFACE]["Name"]
+            for obj, obj_data in objects
+            if StratisDbus._POOL_IFACE in obj_data
+        }
+
+        result = {}
+        for fs_object in fs_objects:
+            created = fs_object["Created"]
+            values = {
```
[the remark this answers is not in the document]
jbaublitz: Resolved.

tests/blackbox/testlib/dbus.py
```diff
@@ -0,0 +1,207 @@
...
+        if objects is None:
+            objects = {}
+
+        pool_objects = {
+            path: obj_data[StratisDbus._POOL_IFACE]
+            for path, obj_data in objects
+            if StratisDbus._POOL_IFACE in obj_data
+        }
+
+        pool_object_path = [
```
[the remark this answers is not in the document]
jbaublitz: Resolved.

tests/blackbox/testlib/dbus.py
```diff
@@ -0,0 +1,207 @@
...
+        return object_manager.GetManagedObjects()
+
+    @staticmethod
+    def pool_list():
+        """
+        Query the pools
+        :return: A list of pool names.
+        """
+        objects = StratisDbus._get_managed_objects().items()
+        if objects is None:
```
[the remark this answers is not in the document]
jbaublitz: Resolved. I think this was lingering from a bug that I incorrectly attributed to `GetManagedObjects()` returning `None`.

tests/blackbox/testlib/dbus.py
```diff
@@ -0,0 +1,187 @@
...
+        )
+        iface.DestroyPool(pool_object_paths[0])
+
+        return None
+
+    @staticmethod
+    def fs_list():
+        """
+        Query the file systems
+        :return: A dict,  Key being the fs name, the value being a dict with
```
mulkieran: Please fix up header comment for updated implementation.

tests/blackbox/testlib/dbus.py
```diff
@@ -0,0 +1,187 @@
...
+    @staticmethod
+    def destroy_all():
+        """
+        Destroys all Stratis FS and pools!
+        :return: None
+        """
+        umount_mdv()
+
+        # Remove FS
+        for name, filesystem in StratisDbus.fs_list().items():
```
mulkieran: ```filesystem``` -> ```poolname``` and change line below also to match.

jbaublitz:
>     * We recently discussed how the StratisCli class filters out all the filesystems and pools that do not start with the distinguishing prefix. Should we be doing that here as well?
> 

That does make sense here. If we want it to be a drop in replacement, it should have the same behavior as well. I'll add that in another commit.

>     * Should we be eliminating the StratisCli class entirely and using the StratisDbus class for clean up both for the daemon tests and for the CLI tests?
> 

I would personally vote for keeping StratisCli for CLI tests because this is a great way to programmatically expose commands and their outputs for inspection in the test suite. However, I do think we should switch over to StratisDbus for tests and set up/tear down alike for the daemon tests.  Ultimately our goal seems to be removing the CLI dependency from the daemon tests so it makes sense to use StratisCli for CLI testing and StratisDbus for daemon testing. Thoughts?

>     * Recall that we had some problems w/ the timeout being too small for some tests the last time. In particular we had to drop the snapshot test because it took too long. We should definitely be setting the timeout value here somewhere, otherwise we'll get the libdbus default (too small!), but the question is how and for how long?

I did some digging and the `Interface` class does some nice things for us here. Similarly to [`call_blocking()`](https://dbus.freedesktop.org/doc/dbus-python/dbus.connection.html#dbus.connection.Connection.call_blocking), any method called on an `Interface` object will accept a keyword argument of `timeout` in the same format as `call_blocking()` meaning that it is a `float` in units of seconds. I have tested the timeout value to make sure it's used and not simply discarded and it does indeed trigger a timeout when set too low. I can add in a keyword argument `timeout` for all of these library methods so that we can provide variable timeouts depending on the test.

tests/blackbox/testlib/dbus.py
jbaublitz: Resolved.

tests/blackbox/testlib/dbus.py
jbaublitz: Resolved.

jbaublitz:
Timeout is now adjustable via the `STRATIS_DBUS_TIMEOUT` environment variable. `StratisDbus` also checks that resources start with `TEST_PREF`. I made one slight modification to the location of `TEST_PREF`, `p_n` and `fs_n` as they are now shared across `StratisCli` and `StratisDbus` so those are now factored out into `utils.py`.

tests/blackbox/testlib/dbus.py
```diff
@@ -0,0 +1,232 @@
...
+            and obj_data[StratisDbus._FS_IFACE]["Name"].startswith(TEST_PREF)
+        ]
+
+        pool_path_to_name = {
+            obj: obj_data[StratisDbus._POOL_IFACE]["Name"]
+            for obj, obj_data in objects
+            if StratisDbus._POOL_IFACE in obj_data
+        }
+
+        result = {
```
mulkieran: Please just return the value here, as ```return {...}```. I like to avoid coming up with identifier names for expressions that are used only once, as I think they clutter up the code.

tests/blackbox/testlib/dbus.py
```diff
@@ -0,0 +1,232 @@
...
+        fs_objects = {
+            path: obj_data[StratisDbus._FS_IFACE]
+            for path, obj_data in objects
+            if StratisDbus._FS_IFACE in obj_data
+        }
+
+        pool_object_paths = [
+            path
+            for path, pool_obj in pool_objects.items()
+            if pool_obj["Name"] == pool_name
```
mulkieran: Might as well match ```TEST_PREF``` here, for consistency.

tests/blackbox/testlib/dbus.py
jbaublitz: Resolved.

tests/blackbox/testlib/dbus.py
jbaublitz: I've made some changes all around in the way that I do this in the most recent commit. I've started filtering in all cases on objects from `GetManagedObjects` before they're used anywhere. I think this is probably best to do right at the beginning so any code that gets added and needs to use those objects in testing doesn't need to worry about filtering independently. Resolved.

PR #346, opened 2019-09-13
Make _stratisd_constants.py have 100% code coverage

mulkieran:


src/stratis_cli/_stratisd_constants.py
```diff
@@ -41,7 +41,11 @@ def the_func(num, terse_unknown=False):
         """
         try:
             the_str = str(klass(num)).rsplit(".")[-1].replace("_", " ")
-        except ValueError:
+
+        # This branch is taken only if the constants defined here do not
+        # match those defined in stratisd. It is not worth gettng coverage
+        # for as we should remedy such a situation very rapidly.
```
bgurney-rh: "It is not worth getting coverage for as we should remedy such a situation very rapidly."

I'm guessing this should be the following (which may require some line re-wrapping):

"It is not worth getting coverage for this as we should remedy such a situation very rapidly."

Issue #336, opened 2019-09-04
Interpret stack trace upon timeout expiration


waltdisgrace:
When the timeout variable is set to a too-small value for a command, a DBusException is raised and the following stack traces are displayed. Instead, we should handle the exception and more concisely notify the user that a timeout expiration has occurred. 

```
[gchin@to-be-determined stratis-cli]$ PYTHONPATH=./src STRATIS_DBUS_TIMEOUT="0" ./bin/stratis pool list
stratis encountered an unexpected error during execution. Please report the error and include in your report the stack trace shown below. 

Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/dbus_python_client_gen/_invokers.py", line 319, in dbus_func
    return dbus_method(*xformed_args, timeout=timeout)
  File "/usr/local/lib64/python3.7/site-packages/dbus/proxies.py", line 145, in __call__
    **keywords)
  File "/usr/local/lib64/python3.7/site-packages/dbus/connection.py", line 651, in call_blocking
    message, timeout)
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NoReply: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/gchin/stratis-storage/stratis-cli/src/stratis_cli/_main.py", line 43, in the_func
    result.func(result)
  File "/home/gchin/stratis-storage/stratis-cli/src/stratis_cli/_actions/_top.py", line 71, in list_pools
    managed_objects = ObjectManager.Methods.GetManagedObjects(proxy, {})
  File "/usr/lib/python3.7/site-packages/dbus_python_client_gen/_invokers.py", line 327, in dbus_func
    name, xformed_args)) from err
dbus_python_client_gen._errors.DPClientInvocationError: Error while invoking method "GetManagedObjects" belonging to interface "org.freedesktop.DBus.ObjectManager" with arguments ()

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "./bin/stratis", line 36, in <module>
    main()
  File "./bin/stratis", line 32, in main
    return run()(sys.argv[1:])
  File "/home/gchin/stratis-storage/stratis-cli/src/stratis_cli/_main.py", line 59, in the_func
    handle_error(err)
[...]

PR #347, opened 2019-09-16
Interpret stack trace upon timeout expiration

waltdisgrace:
Resolves: #336.

### DIFF
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -146,6 +146,20 @@ def interpret_errors(errors):
         ):
             return "Most likely stratis is unable to connect to the stratisd D-Bus service."
 
+        # pylint: disable=fixme
+        # FIXME: remove no coverage pragma when adequate testing for CLI
+        # output exists.
+        if (
+            # pylint: disable=bad-continuation
+            isinstance(error, dbus.exceptions.DBusException)
+            and error.get_dbus_name() == "org.freedesktop.DBus.Error.NoReply"
+        ):  # pragma: no cover
+            fmt_str = (
+                "stratis attempted communication with the daemon, stratisd, "
+                "over the D-Bus, but stratisd did not respond in the allowed time."
+            )
+            return fmt_str % error
+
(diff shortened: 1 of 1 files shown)


src/stratis_cli/_error_reporting.py
```diff
@@ -145,7 +145,13 @@ def interpret_errors(errors):
             and error.get_dbus_name() == "org.freedesktop.DBus.Error.NameHasNoOwner"
         ):
             return "Most likely stratis is unable to connect to the stratisd D-Bus service."
-
+        if (
+            # pylint: disable=bad-continuation
+            isinstance(error, dbus.exceptions.DBusException)
+            and error.get_dbus_name() == "org.freedesktop.DBus.Error.NoReply"
+        ):  # pragma: no cover
+            return """Stratis attempted communication with the daemon, stratisd,
```
mulkieran: Please use the approach for multi-line strings that is used at line 120 in this file.

mulkieran: Use lowercase for Stratis even though it starts the sentence. We want to be clear that we are talking about the CLI tool, not the project.

src/stratis_cli/_error_reporting.py
```diff
@@ -145,7 +145,13 @@ def interpret_errors(errors):
...
+            over the D-Bus, but stratisd did not respond in the allowed time."""
```
mulkieran: Prefer an empty line in between this and the next section.

waltdisgrace:
rebased.

src/stratis_cli/_error_reporting.py
```diff
@@ -145,6 +145,16 @@ def interpret_errors(errors):
             and error.get_dbus_name() == "org.freedesktop.DBus.Error.NameHasNoOwner"
         ):
             return "Most likely stratis is unable to connect to the stratisd D-Bus service."
+        if (
```
mulkieran: Please go ahead and add the three line FIXME, like with the other parts of this method that we are not testing.

PR #350, opened 2019-09-20
Add Stratis CLI blackbox tests

waltdisgrace:
Addresses:  https://github.com/stratis-storage/project/issues/79

The following tests have been added:

- test_stratisd_redundancy
- test_pool_list_empty
- test_filesystem_list_empty
- test_pool_create
- test_pool_list_not_empty
- test_blockdev_list
- test_pool_create_same_name
- test_pool_add_cache
- test_pool_destroy
- test_filesystem_create
- test_pool_add_data
- test_filesystem_list_not_empty
- test_filesystem_create_same_name
- test_filesystem_rename
- test_filesystem_rename_same_name
- test_filesystem_snapshot
- test_filesystem_destroy

### DIFF
--- a/tests/blackbox/stratis_cli_cert.py
+++ b/tests/blackbox/stratis_cli_cert.py
@@ -82,7 +82,6 @@ class StratisCertify(unittest.TestCase):
     ):
         """
         Execute a test command and make assertions about the exit code, stderr, and stdout
-        :param self: A test object
         :param list args: The arguments needed to execute the Stratis command being tested
         :type args: List of str
         :param exp_exit_code: The expected exit code, 0, 1, or 2
--- a/tests/blackbox/stratis_cli_cert.py
+++ b/tests/blackbox/stratis_cli_cert.py
@@ -39,7 +39,7 @@ def make_test_pool():
 def make_test_filesystem(pool_name):
     """
     Create a test filesystem that will later get destroyed
-    :param name: Name of a test pool
+    :param pool_name: Name of a test pool
     :return: Name of the created filesystem
     """
     filesystem_name = fs_n()
@@ -50,6 +50,45 @@ def make_test_filesystem(pool_name):
         == 0
     )
     return filesystem_name
+
+
+def assert_test_result_success_stdout_empty(self, exit_code, stdout, stderr):
+    """
+    :param self: A test object
+    :param exit_code: An exit code
+    :param stdout: stdout
+    :param stderr: stderr
+    :return: None
+    """
+    self.assertEqual(exit_code, 0)
+    self.assertEqual(stderr, "")
+    self.assertEqual(stdout, "")
+
+
+def assert_test_result_success_stdout_nonempty(self, exit_code, stdout, stderr):
+    """
+    :param self: A test object
+    :param exit_code: An exit code
+    :param stdout: stdout
+    :param stderr: stderr
+    :return: None
+    """
+    self.assertEqual(exit_code, 0)
+    self.assertEqual(stderr, "")
+    self.assertNotEqual(stdout, "")
+
+
+def assert_test_result_failure_stdout_empty(self, exit_code, stdout, stderr):
+    """
+    :param self: A test object
+    :param exit_code: An exit code
+    :param stdout: stdout
+    :param stderr: stderr
+    :return: None
+    """
+    self.assertEqual(exit_code, 1)
+    self.assertNotEqual(stderr, "")
+    self.assertEqual(stdout, "")
 
 
 class StratisCertify(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


tests/blackbox/stratis_cli_cert.py
```diff
@@ -20,12 +20,77 @@
...
 DISKS = []
 
 
+def make_test_pool():
+    """
+    Create a test pool that will later get destroyed
+    :return: Name of the created pool
+    """
+    pool_name = p_n()
+    exec_test_command([STRATIS_CLI, "pool", "create", pool_name, DISKS[0]])
```
mulkieran: Better to assert here too, like w/ ```make_test_filesystem```.

tests/blackbox/stratis_cli_cert.py
```diff
@@ -20,12 +20,77 @@
...
+    assert (
+        exec_test_command(
+            [STRATIS_CLI, "filesystem", "create", pool_name, filesystem_name]
+        )[0]
+        == 0
+    )
+    return filesystem_name
+
+
+def assert_test_result_success_stdout_empty(self, exit_code, stdout, stderr):
```
mulkieran: Two comments on these abstracted testing methods:

1. It probably makes sense to pass the command line as an argument also.

2. You can probably collapse these methods all into a single method more efficiently. I think originally I thought there were only 8 possibilities, but actually there are 12. We have 3 possibilities for the error code: 0 (when things go well), 2 (when there is an error in parsing the command line arguments), 1 (for all other errors). For stdout and stderr we're trying to go w/ distinguishing only between empty and non-empty. This gives 3 * 2 * 2 = 12 possible choices. We don't want 12 methods distinguished by their names, so it's probably better to have one method w/ parameters that govern its behavior.

In your scheme, you are already passing the expected exit code. This means that you would expect to have 12 / 3 = 4 distinct methods. But it would really be easier and more extensible for the user if you used, for example, boolean arguments to indicate whether stderr and stdout were expected to be empty or non-empty.

tests/blackbox/stratis_cli_cert.py
```diff
@@ -20,12 +20,66 @@
...
+    assert (
+        exec_test_command(
+            [STRATIS_CLI, "filesystem", "create", pool_name, filesystem_name]
+        )[0]
+        == 0
+    )
+    return filesystem_name
+
+
+def test_command(  # pylint: disable=bad-continuation
```
mulkieran: Rename ```test_command``` to something that doesn't start with ```test``` and move the implementation inside the ```StratisCertify``` class definition. It's still an instance method of the class, which is why you pass it ```self``` so it would typically be defined in the class. But unittest likes to interpret all methods that start with ```test``` as representing tests, if they find it in a class that extends ```TestCase```, so you'll want to name the method something else to avoid that.

Then, wherever you call the method, use the ```self.test_command(...)``` syntax instead of the syntax you're using currently, ```test_command(self, ...)```.

tests/blackbox/stratis_cli_cert.py
```diff
@@ -20,12 +20,66 @@
...
+    self, command_line_args, exp_exit_code, exp_stderr_is_empty, exp_stdout_is_empty
+):
+    """
+    Execute a test command and make assertions about the exit code, stderr, and stdout
+    :param self: A test object
+    :param command_line_args: A list of command line arguments needed to execute the Stratis command
+    :type command_line_args: List of str
+    :param exp_exit_code: The expected exit code, 0, 1, or 2
+    :param exp_stdout_is_empty: A boolean representing whether or not we expect stdout to be empty
```
mulkieran: Make it easier for the users to figure out what these two parameters mean:

```
:param bool exp_stdout_is empty: True if sdout is expected to be empty, otherwise False
```

Same for the other one.

Note that the doc comments are in reverse order with the order of the parameters, too. That could really confuse a reader.

tests/blackbox/stratis_cli_cert.py
mulkieran: Better to assert here too, like w/ ```make_test_filesystem```.

tests/blackbox/stratis_cli_cert.py
```diff
@@ -49,16 +77,219 @@ def setUp(self):
         StratisCli.destroy_all()
         assert StratisCli.pool_list() == []
 
+    def unittest_command(  # pylint: disable=bad-continuation
+        self, args, exp_exit_code, exp_stderr_is_empty, exp_stdout_is_empty
+    ):
+        """
+        Execute a test command and make assertions about the exit code, stderr, and stdout
+        :param self: A test object
```
mulkieran: Please leave out the doc line for ```self```. It's never documented if the method is part of a class.

mulkieran:
Superceded by #357.

PR #356, opened 2019-10-04
Add stratisd blackbox tests

waltdisgrace:


mulkieran:
Travis error looks pretty real to me.

tests/blackbox/stratisd_cert.py
```diff
@@ -70,6 +99,162 @@ def test_get_managed_objects(self):
...
+        )
+
+    def test_filesystem_snapshot(self):
+        """
+        Test snapshotting a filesystem.
+        """
+        pool_name = make_test_pool()
+        filesystem_name = fs_n()
+        snapshot_name = fs_n()
+        assert StratisDbus.filesystem_create(pool_name, filesystem_name)
```
mulkieran: Please make ```assert```s more specific. You want the assertion to fail if ```filesystem_create``` did not create the specified filesystem, but I think it will succeed regardless. I think this holds for all the ```assert```s in the tests that I can see.

tests/blackbox/stratisd_cert.py
```diff
@@ -20,12 +20,41 @@
...
 DISKS = []
 
 
+def make_test_pool():
+    """
+    Create a test pool that will later get destroyed
+    :return: Name of the created pool
+    """
+    pool_name = p_n()
+    assert (exec_test_command([STRATIS_CLI, "pool", "create", pool_name, DISKS[0]]))[
```
mulkieran: You want to avoid using the CLI anywhere in these stratisd tests. So you'll want to find a different way to define ```make_test_pool``` and ```make_test_filesystem``` which uses ```StratisDbus``` methods.

mulkieran:
superceded by #370 .

PR #358, opened 2019-10-10
Convert no-op engine results into failures

jbaublitz:
Supersedes #321 
Closes stratis-storage/project#51

Leverage stratisd changes to determine when actions have had an effect and fail on no effect.

PR #361, opened 2019-10-11
blackbox: change cli fs_rename_same_name test to verify for failure

bgurney-rh:
The test_filesystem_rename_same_name test should now verify that
renaming to the same name fails, since the failure return code
on the command line resulted in no action.


bgurney-rh:
The stratis-cli test failure is most likely due to #358 needing to be merged prior to this.  I can rebase it afterward.

bgurney-rh:
Rebased.

PR #370, opened 2019-10-22
 Add stratisd blackbox tests

waltdisgrace:


tests/blackbox/stratisd_cert.py
```diff
@@ -20,12 +20,44 @@
...
+    Create a test filesystem that will later get destroyed
+    :param pool_name: Object path of a test pool
+    :return: Object path of the created filesystem or None
+    """
+    (
+        filesystems_created,
+        (array_of_tuples_with_obj_paths_and_names),
+    ), return_code, _ = StratisDbus.filesystem_create(pool_object_path, filesystem_name)
+    if filesystems_created == dbus.Boolean(False) or return_code == dbus.UInt16(1):
+        print("Failed to create filesystem.")
```
mulkieran: Please omit the ```print```.

tests/blackbox/stratisd_cert.py
```diff
@@ -20,12 +20,44 @@
...
+
 from testlib.dbus import StratisDbus, clean_up
+from testlib.utils import exec_command, exec_test_command, process_exists, p_n, fs_n
 
 DISKS = []
 
 
+def make_test_pool(pool_name):
+    """
+    Create a test pool that will later get destroyed
```
mulkieran: Please document ```param```.

tests/blackbox/stratisd_cert.py
```diff
@@ -20,12 +20,44 @@
...
+    )
+    if obj_path_exists == dbus.Boolean(False) or return_code == dbus.UInt16(1):
+        return None
+    return obj_path
+
+
+def make_test_filesystem(pool_object_path, filesystem_name):
+    """
+    Create a test filesystem that will later get destroyed
+    :param pool_name: Object path of a test pool
```
mulkieran: Param documentation does't match reality here.

tests/blackbox/stratisd_cert.py
```diff
@@ -20,12 +20,44 @@
...
 
 DISKS = []
 
 
+def make_test_pool(pool_name):
+    """
+    Create a test pool that will later get destroyed
+    :return: Object path of the created pool or None
+    """
+    pool_disk = ["/dev/nvme0n1p1"]
```
mulkieran: The line above will not work on the blackbox tests under general conditions. You will have to get this device from which to construct the test pool from the devices passed as arguments to the script.

tests/blackbox/stratisd_cert.py
```diff
@@ -20,12 +20,45 @@
...
 from testlib.dbus import StratisDbus, clean_up
+from testlib.utils import exec_command, exec_test_command, process_exists, p_n, fs_n
 
 DISKS = []
 
 
+def make_test_pool(pool_name):
+    """
+    Create a test pool that will later get destroyed
+    :param pool_name: Name of the pool to be created
```
mulkieran: Please be complete: ```param str pool_name```.

tests/blackbox/stratisd_cert.py
```diff
@@ -20,12 +20,45 @@
...
+    )
+    if obj_path_exists == dbus.Boolean(False) or return_code == dbus.UInt16(1):
+        return None
+    return obj_path
+
+
+def make_test_filesystem(pool_path, fs_name):
+    """
+    Create a test filesystem that will later get destroyed
+    :param pool_path: Object path of a test pool
```
mulkieran: Please be complete here as well, i.e., include the type in the parameter docs.

tests/blackbox/testlib/dbus.py
```diff
@@ -88,6 +89,28 @@ def _get_managed_objects():
...
+        ]
+
+        return versions[0]
+
+    @staticmethod
+    def stratisd_redundancy():
+        """
+        Get stratisd redundancy
+        """
+        redundancy = "NONE = 0"
```
mulkieran: Use ```return "NONE = 0"```, rather than the two lines. Also, please document why this is Ok. It very much looks like your test is a fake, since the test infrastructure is just returning its own value.

tests/blackbox/testlib/dbus.py
```diff
@@ -88,6 +89,28 @@ def _get_managed_objects():
         )
         return object_manager.GetManagedObjects(timeout=StratisDbus._TIMEOUT)
 
+    @staticmethod
+    def stratisd_version():
+        """
+        Get stratisd version
```
mulkieran: Please document the return type here:

```:rtype: <????>```

File: tests/blackbox/testlib/dbus.py [hunk 9]
```diff
@@ -88,6 +89,28 @@ def _get_managed_objects():
...
+            for _, obj_data in StratisDbus._get_managed_objects().items()
+            if StratisDbus._MNGR_IFACE in obj_data
+        ]
+
+        return versions[0]
+
+    @staticmethod
+    def stratisd_redundancy():
+        """
+        Get stratisd redundancy
```
mulkieran: Please document the return type here also and wherever else it is missing.

mulkieran:
@waltdisgrace I've had second thoughts about testing redundancy. I think you might as well just not do it at all, since it doesn't test the daemon at all, and that's our purpose with these tests. So please go ahead and remove the infrastructure to support getting redundancy and also the test for it.

tests/blackbox/stratisd_cert.py
```diff
@@ -70,6 +103,206 @@ def test_get_managed_objects(self):
         self.assertEqual(stderr, "")
         self.assertNotEqual(stdout, "")
 
+    def test_stratisd_version(self):
+        """
+        Test getting the daemon version.
+        """
+        self.assertNotEqual(StratisDbus.stratisd_version(), [])
```
mulkieran: Here's a good example of where types are important. You documented ```stratisd_version()``` as returning a string, and I think that's quite right. But here you're asserting that it's not equal to ```[]``` which is the empty list. A string can not ever be equal to the empty list, so this test will always succeed, so long as stratisd is alive enough to return a properly formatted GetManagedObjects() result. You want something more precise, but not so precise as to be excessively finicky, i.e., you'll definitely want to avoid string matching using regular expressions. For this test, I'ld suggest something like this:

Try using the method ```assertIsInstance```. There's a related method, ```isinstance``` that behaves similarly.:

```
>>> isinstance
<built-in function isinstance>
>>> isinstance("i am a string", str)
True
>>> isinstance(int(32), int)
True
>>> isinstance(int(32), double)
  File "<stdin>", line 1, in <module>
NameError: name 'double' is not defined
>>> isinstance(int(32), float)
False
>>> isinstance(dbus.String("abc"), str)
True
```
and so forth. Use the method ```assertIsInstance``` to assert that the method is an instance of the correct type. Then, use other assertions to check the value. This way, the test cuts much closer to the boundary between correct and not.

For this particular test, I'ld try:

```
result = StratisDbus.stratisd_version()
self.assertIsInstance(result, str)
self.assertIsNotEqual(result, "")
```

waltdisgrace:
rebased.

waltdisgrace:
rebased.

PR #374, opened 2019-10-23
Preliminary work for supporting new propery interface in CLI

jbaublitz:
This is a preliminary PR addressing the CLI side of stratis-storage/project#52. Putting up for initial review.


jbaublitz:
I seem to have found a way to maintain the list comprehensions. I think the code I've just pushed will be far cleaner.

jbaublitz:
@bgurney-rh Just a note that because this is another breaking API change, the CI will fail until we merge the stratisd changes. I'm going to run the tests manually and all CI failures in this PR can safely be ignored.

src/stratis_cli/_actions/_data.py
```diff
@@ -256,6 +268,11 @@ def _get_timeout(value):
         environ.get("STRATIS_DBUS_TIMEOUT", DBUS_TIMEOUT_SECONDS * 1000)
     )
 
+    fetch_properties_spec = ET.fromstring(SPECS[FETCH_PROPERTIES_INTERFACE])
+    FetchProperties = make_class("FetchProperties", fetch_properties_spec, timeout)
+    MOFetchProperties = managed_object_class("MOFetchProperties", fetch_properties_spec)
```
mulkieran: Omit these two lines, they're only used for getting objects that have properties out of ```GetManagedObjects``` result, so I don't think we will be using the generated classes ever.

src/stratis_cli/_actions/_data.py
```diff
@@ -210,6 +198,30 @@
 MAXIMUM_DBUS_TIMEOUT_MS = 1073741823
 
 
+def _fetch_property(name, props, to_string):
```
mulkieran: Please move this functionality into ```formatting.py``` as right now there is no necessity for stringifying these values other than for formatting.

Please also adapt the function so that it generates values suitable for printing in a columnar format when there is an error or the key is not found. You definitely don't want to print out the error from the daemon in the column; we can't guarantee that it is not enormous.

Ideally, you can parameterize the behavior for key not found or error just as you are already parameterizing for the value.

This would be easy to do with a subclass of dict, or alternatively a lighter-weight currying approach.

mulkieran: P.S. There may be some documentation in the D-Bus API doc about how unknown, etc things are supposed to be printed out. If so, we should try to follow it, or if it seems wrong, revise it.

mulkieran: I notice that in ```_stratisd_constants.py``` there seems to be a value "???".

jbaublitz: I've made a second pass at this.

src/stratis_cli/_actions/_data.py
jbaublitz: Resolved.

jbaublitz:
I've made a second pass at incorporating all of the feedback. Putting this back in review.

src/stratis_cli/_errors.py
```diff
@@ -33,6 +33,32 @@ class StratisCliRuntimeError(StratisCliError):
...
+            :param str type_name: the name of the type that does not support this property name
+            :param str prop_name: the property that did not exist
+        """
+        # pylint: disable=super-init-not-called
+        self.type_name = type_name
+        self.prop_name = prop_name
+
+    # pylint: disable=fixme
+    # FIXME: Find a way to test the string conversion method. This will
+    # require a call directly to the DBus interface.
```
mulkieran: Instead of putting ```no cover``` pragma here, consider putting it on the whole class. See ```StratisCliUnkownInterfaceError``` for an example. I think that the reasoning would be kind of the same.

jbaublitz: Resolved.

src/stratis_cli/_actions/_formatting.py
```diff
@@ -38,6 +40,33 @@
     maybe_wcswidth = len  # pragma: no cover
 
 
+def _fetch_property(object_type, props, name, to_string):
```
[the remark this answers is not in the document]
jbaublitz: Resolved.

src/stratis_cli/_errors.py
```diff
@@ -33,6 +33,31 @@ class StratisCliRuntimeError(StratisCliError):
...
 
+# This indicates a bug and so coverage is not necessary as this error
+# is not handled.
+class StratisCliPropertyNotFoundError(StratisCliRuntimeError):  # pragma: no cover
+    """
+    Exception raised when a requested property from FetchProperties DBus interface
+    does not exist.
+    """
+
+    def __init__(self, type_name, prop_name):
```
mulkieran: I'ld prefer the type name to be the name of the _D-Bus interface_ to which the object belongs. That is sort of a user unfriendly thing to report, but because we expect this to be a bug if it occurs, and since that is the precise identifier for the thing that we have failed to get the property for if we raise the exception, that seems the best choice. Any other choice would be kind of vague.

jbaublitz: Oops! Sorry, you already requested this. Doing this now.

jbaublitz: I missed this in my last round of revisions.

jbaublitz:
I've made the last revision. I'm running tests manually now given the CI mismatch.

src/stratis_cli/_errors.py
jbaublitz: Resolved.

jbaublitz:
I've run the tests against the final version of each PR. Assuming there are no more changes needed, CI should pass once both are merged.

jbaublitz:
Oops, caught one more thing. I'm going to make one more change: move string literals over to the const values already defined.

src/stratis_cli/_actions/_logical.py
```diff
@@ -98,6 +98,7 @@ def list_volumes(namespace):
         List the volumes in a pool.
         """
         # pylint: disable=import-outside-toplevel
+        from ._constants import FILESYSTEM_INTERFACE
```
mulkieran: Please move this import up to top of file, with other imports not from ```_data.py```, and same with similar.

src/stratis_cli/_actions/_logical.py
```diff
@@ -98,11 +98,14 @@ def list_volumes(namespace):
...
+        from ._data import FetchProperties
         from ._data import MOFilesystem
         from ._data import MOPool
         from ._data import ObjectManager
         from ._data import filesystems
         from ._data import pools
+        from ._formatting import fetch_property
```
mulkieran: Here's another import that I'ld like also like  to be at the top...completely missed that the last time.

src/stratis_cli/_errors.py
```diff
@@ -33,6 +33,31 @@ class StratisCliRuntimeError(StratisCliError):
...
+        """ Initializer.
+
+            :param str type_name: the name of the type that does not support this property name
```
mulkieran: Please update docs and error string since using the interface name. Might as well change ```type_name``` to ```interface_name``` here as well.

PR #379, opened 2019-10-29
Add stratisd blackbox tests

waltdisgrace:
Resolves: https://github.com/stratis-storage/project/issues/80
Supercedes: #370

### DIFF
--- a/tests/blackbox/stratisd_cert.py
+++ b/tests/blackbox/stratisd_cert.py
@@ -20,10 +20,41 @@ import sys
 import time
 import unittest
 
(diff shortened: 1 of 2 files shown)


mulkieran:
@waltdisgrace I think it's reasonable to put this up for final review. There is one thing that I'ld like changed, but it is in the CLI tests also, as I must have overlooked it when doing review of that PR. So I put up a separate issue for that: #380, and I'll assign that to you to be worked on as soon as this is merged.

waltdisgrace:
rebased.

Issue #380, opened 2019-10-29
Don't hard-code the disks for making the test pools in blackbox tests scripts


mulkieran:
It would be better not to hard code the list of pool disks in ```make_test_pool```. Instead, try passing a list of disks as a parameter.

The reason for this is very practical. We have known instances of pathological cases where the cache was too big for its data devices. I forget what actually happened, but in normal situations the total size of the cache devices should always be less than the total size of the data devices. So we want the flexibility to pick an arbitrary subset of devices from DISKS in order to build the pool, depending on what we're actually testing. The clearest way is to pass a list to this method and supply it as an argument in the places where it is called.

After that, it'll probably be necessary to revise a couple of tests based on their meaning. For example, going back to caches, in the cache test it would be better to build the pool from two devices, and then make the cache from a single device, just to get the proportions correct.

mulkieran:
Closed by #403.

PR #382, opened 2019-10-30
Pass disks for test pool creation as a parameter

waltdisgrace:
Addresses #380.

### DIFF
--- a/tests/blackbox/stratis_cli_cert.py
+++ b/tests/blackbox/stratis_cli_cert.py
@@ -26,9 +26,10 @@ from testlib.stratis import STRATIS_CLI, clean_up
 DISKS = []
 
 
-def make_test_pool():
+def make_test_pool(pool_disks):
     """
     Create a test pool that will later get destroyed
+    :param list pool_disks: List of disks with which the pool will be created
     :return: Name of the created pool
     """
     pool_name = p_n()
@@ -142,7 +143,7 @@ class StratisCertify(unittest.TestCase):
         """
         Test listing an existent pool.
         """
-        make_test_pool()
+        make_test_pool(DISKS[0])
         self.unittest_command([STRATIS_CLI, "pool", "list"], 0, True, False)
 
     def test_blockdev_list(self):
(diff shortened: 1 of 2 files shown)


tests/blackbox/stratisd_cert.py
```diff
@@ -28,15 +28,15 @@
 DISKS = []
 
 
-def make_test_pool(pool_name):
+def make_test_pool(pool_name, pool_disks):
     """
     Create a test pool that will later get destroyed
     :param str pool_name: Name of the pool to be created
+    :param list disks: List of disks with which the pool will be created
```
mulkieran: Just a minor typo: ^disks^pool_disks

tests/blackbox/stratis_cli_cert.py
```diff
@@ -154,23 +155,29 @@ def test_pool_create_same_name(self):
         Test creating a pool that already exists.
         """
         self.unittest_command(
-            [STRATIS_CLI, "pool", "create", make_test_pool(), DISKS[0]], 1, False, True
+            [STRATIS_CLI, "pool", "create", make_test_pool(DISKS[0]), DISKS[1]],
+            0,
```
mulkieran: Typo here...compare with the previous code and you'll see that your expectations of the results of the test have changed, which shouldn't happen.

tests/blackbox/stratis_cli_cert.py
```diff
@@ -140,7 +141,7 @@ def test_pool_list_not_empty(self):
         """
         Test listing an existent pool.
         """
-        make_test_pool()
+        make_test_pool(DISKS[0])
```
mulkieran: Make sure to always use a slice when selecting from the DISKS array, as you do in stratisd_cert.py. A slice of a list of strings is itself a list of strings, but an element of a list of strings is just a string.

tests/blackbox/stratis_cli_cert.py
```diff
@@ -26,13 +26,14 @@
...
-def make_test_pool():
+def make_test_pool(pool_disks):
     """
     Create a test pool that will later get destroyed
+    :param list pool_disks: List of disks with which the pool will be created
     :return: Name of the created pool
     """
     pool_name = p_n()
-    assert (exec_test_command([STRATIS_CLI, "pool", "create", pool_name, DISKS[0]]))[
+    assert (exec_test_command([STRATIS_CLI, "pool", "create", pool_name, pool_disks]))[
```
mulkieran: The place to look for the bug is just here. You documented ```pool_disks``` to be a list of str, essentially, but you use it as if it a str. It is correct to call ```pool_disks``` a list of str. So the error lies in the fact that the code within the method treats ```pool_disks``` as if it were a str.

PR #383, opened 2019-10-30
Expect bytes from size methods in stratisd

jbaublitz:
Change CLI to expect bytes from size calls to stratisd.

These are the client side changes for stratis-storage/stratisd#1243.


jbaublitz:
@bgurney-rh I don't think tests will fail. They may, however, and if they do, we will do the same thing as the last PR and merge stratisd changes then retest.

PR #389, opened 2019-10-31
Change DBus path stratis1 to stratis2 in preparation for new version

jbaublitz:
Closes stratis-storage/project#103

### DIFF
--- a/tests/blackbox/stratisd_cert.py
+++ b/tests/blackbox/stratisd_cert.py
@@ -23,7 +23,7 @@ import unittest
 import dbus
 
 from testlib.dbus import StratisDbus, clean_up
-from testlib.utils import exec_command, exec_test_command, process_exists, p_n, fs_n
+from testlib.utils import exec_command, process_exists, p_n, fs_n
 
 DISKS = []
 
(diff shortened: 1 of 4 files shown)


mulkieran:
@jbaublitz Please rebase when you get the chance.

tests/blackbox/stratisd_cert.py
```diff
@@ -90,8 +90,8 @@ def test_get_managed_objects(self):
             [
                 "busctl",
                 "call",
-                "org.storage.stratis1",
-                "/org/storage/stratis1",
+                "org.storage.stratis2",
```
mulkieran: Please use constants defined in testlib/dbus.py here, instead of literal strings and make them public.

mulkieran: We realized that we have a fine method for doing this already in testlib/dbus.py. So we'll call that method, instead, and these string literals will just go away as a result.

jbaublitz: I've switched `_get_managed_objects()` to `get_managed_objects()` and am using that instead of busctl. Resolved.

jbaublitz:
@mulkieran @bgurney-rh Rebased and ready to go.

PR #393, opened 2019-11-01
Simplify StratisCli class

mulkieran:


jbaublitz:
@bgurney-rh @mulkieran I'm looking at this and do we want to stop checking for the test prefix in `destroy_all()`? This could potentially destroy other pools and filesystems that are not marked as testing resources.

mulkieran:
> @bgurney-rh @mulkieran I'm looking at this and do we want to stop checking for the test prefix in `destroy_all()`? This could potentially destroy other pools and filesystems that are not marked as testing resources.

I will add an additional commit that should remove that concern entirely.

mulkieran:
@jbaublitz Note that the StratisCli list methods already exclude all test filesystems or pools that have not been made w/ the test prefix. If we make the StratisCli class private, then nobody should be able to use the class for anything other than infrastructure cleanup, so its function should be clearer.

PR #394, opened 2019-11-01
Call "udevadm settle" once setup for each individual test is completed

mulkieran:


mulkieran:
@bgurney-rh Can you test this PR before it is merged? It would be pointless to merge it if it does not fix the problem.

mulkieran:
rebased.

PR #398, opened 2019-11-04
Update changelog prior to new release

mulkieran:

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,11 +4,37 @@ Python auto-formatter: black (18.9b0)
 
 New minimum stratisd version requirement: 2.0.0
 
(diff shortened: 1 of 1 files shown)


mulkieran:
Note that some things that would be under "Tidies and Maintenance" if they were deferred until the next release, are here put under the change that they are closely associated with, since they go in in this release. For example, removing dead code generally belongs in "Tidies and Maintenance", but if it was dead code introduced by one of the significant changes called out in this version, then it goes with that change instead.

mulkieran:
The CLI has been busier than the daemon, so I'm going to go ahead and put this up for review right now. There should be one more update before it finally goes in.

PR #406, opened 2019-11-06
Display PhysicalSize triple for pool list

mulkieran:


mulkieran:
Sample output:

```
$ PYTHONPATH=./src ./bin/stratis pool list
Name               Total Physical
poolname  512 EiB / 0 B / 512 EiB
$ PYTHONPATH=./src ./bin/stratis filesystem list
Pool Name  Name    Used       Created            Device                    UUID                            
poolname   fsname  11.77 MiB  Nov 06 2019 11:06  /stratis/poolname/fsname  738b55bb243448d0bd53d9ff21bcddfa
$ PYTHONPATH=./src ./bin/stratis blockdev list
Pool Name  Device Node  Physical Size  Tier
poolname   /dev/jkjl            1 GiB  Data
poolname   /dev/jlk             1 GiB  Data
```

This is using the sim engine, so pool total size and used are just set arbitrarily. I'll try another with more interesting numbers.

mulkieran:
Jenkins failure is just as should be expected.

mulkieran:
Example showing the small size, large size problem, and that justbytes solves it here, well:

```
$ PYTHONPATH=./src ./bin/stratis pool list
Name                       Total Physical
poolname  512 EiB / 512.00 EiB / 2.50 MiB
```

pool is very near full, but not quite. The observable difference between used and total at the given precision is 0, but the actual difference is 2.50 MiB. I know this is an exaggerated example, in this case.

mulkieran:
rebased.

Issue #384, opened 2019-10-30
We should run up some tests to create objects of each error type and stringify them


mulkieran:
The purpose would be just to make sure  that the string formatting wasn't broken in some way that pylint couldn't identify.

There are exceptions that we can't raise through normal testing of our code, but we would like them not to fail when they do get printed out.

And, we can lose a few "no cover" annotations w/out too much trouble, making our code that little bit tidier.

waltdisgrace:
Closed by #411.

PR #408, opened 2019-11-07
Add error type string formatting tests

waltdisgrace:
Addresses #384.

### DIFF
--- a/src/stratis_cli/_errors.py
+++ b/src/stratis_cli/_errors.py
@@ -35,7 +35,7 @@ class StratisCliRuntimeError(StratisCliError):
 
 # This indicates a bug and so coverage is not necessary as this error
 # is not handled.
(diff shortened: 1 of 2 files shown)


jbaublitz:
@waltdisgrace I'm going to hold off on merging this until @mulkieran gets back. I'm not sure whether @mulkieran wanted to simply test that `__str__` doesn't fail in some kind of obviously spectacular way such as with an exception or if we're really looking to check that the string renders the way we want it to. If it's the former, this approach works. If it's the latter, we're only checking that some string is returned. If that's the case, ideally I would do something along the lines of checking that whatever parameters are passed into the error constructor appear in the final rendered string, but we should ultimately go with what she suggests.

bgurney-rh:
This PR was set up from the wrong branch; we'll retry with a new PR.

PR #409, opened 2019-11-07
Add error type string formatting tests

waltdisgrace:
Addresses #384.
Supersedes #408.

### DIFF
--- a/tests/whitebox/unittest/test_error_fmt.py
+++ b/tests/whitebox/unittest/test_error_fmt.py
@@ -1,0 +1,128 @@
+# Copyright 2016 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Test error type string formatting.
+"""
+
+import argparse
+import unittest
+
+from stratis_cli._errors import StratisCliError
+from stratis_cli._errors import StratisCliRuntimeError
+from stratis_cli._errors import StratisCliPropertyNotFoundError
+from stratis_cli._errors import StratisCliPartialChangeError
+from stratis_cli._errors import StratisCliNoChangeError
+from stratis_cli._errors import StratisCliIncoherenceError
+from stratis_cli._errors import StratisCliInUseError
+from stratis_cli._errors import StratisCliUnknownInterfaceError
+from stratis_cli._errors import StratisCliEngineError
+from stratis_cli._errors import StratisCliActionError
+from stratis_cli._errors import StratisCliGenerationError
+from stratis_cli._errors import StratisCliEnvironmentError
+
+from stratis_cli._stratisd_constants import BlockDevTiers
+
+
+class ErrorFmtTestCase(unittest.TestCase):
+    """
+    Test stringification of various error types.
+    """
+
+    def testStratisCliErrorFmt(self):
+        """
+        Test 'StratisCliError'
+        """
+        exception = StratisCliError("Error")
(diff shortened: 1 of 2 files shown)


jbaublitz:
Holding off on this for now. See [here](https://github.com/stratis-storage/stratis-cli/pull/408#issuecomment-551276158).

tests/whitebox/unittest/test_error_fmt.py
```diff
@@ -0,0 +1,128 @@
...
+    def testStratisCliNoChangeErrorFmt(self):
+        """
+        Test 'StratisCliNoChangeError'
+        """
+        exception = StratisCliNoChangeError("Command", frozenset("ChangedResources"))
+        self.assertNotEqual(str(exception), "")
+
+    def testStratisCliIncoherenceErrorFmt(self):
+        """
+        Should fail because the timeout value is a float.
```
jbaublitz: This looks like it may be a copy paste error.

tests/whitebox/unittest/test_error_fmt.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        exception = StratisCliIncoherenceError("Error")
+        self.assertNotEqual(str(exception), "")
+
+    def testStratisCliInUseErrorFmt(self):
+        """
+        Should fail because the timeout value is not an integer.
```
jbaublitz: This looks like it may be a copy paste error as well.

tests/whitebox/unittest/test_error_fmt.py
```diff
@@ -0,0 +1,128 @@
...
+from stratis_cli._errors import StratisCliUnknownInterfaceError
+from stratis_cli._errors import StratisCliEngineError
+from stratis_cli._errors import StratisCliActionError
+from stratis_cli._errors import StratisCliGenerationError
+from stratis_cli._errors import StratisCliEnvironmentError
+
+from stratis_cli._stratisd_constants import BlockDevTiers
+
+
+class ErrorFmtTestCase(unittest.TestCase):
```
jbaublitz: Given how much boilerplate code is in these tests, I would recommend factoring out all of the checking into a separate function. That way you can pass in the exception as an argument and do the same check on them all.

tests/whitebox/unittest/test_error_fmt.py
```diff
@@ -0,0 +1,128 @@
+# Copyright 2016 Red Hat, Inc.
```
jbaublitz: `# Copyright 2019 Red Hat, Inc.` as discussed.

tests/whitebox/unittest/test_error_fmt.py
```diff
@@ -0,0 +1,133 @@
...
+
+class ErrorFmtTestCase(unittest.TestCase):
+    """
+    Test stringification of various error types.
+    """
+
+    def StringNotEmpty(self, exception):
+        """
+        :param exception: an object of a Stratis CLI error type
+        :param type exception: the class defining the Stratis CLI error type
```
jbaublitz: Please fix the doc string. See [here](https://github.com/stratis-storage/stratis-cli/blob/develop/src/stratis_cli/_errors.py#L73) for single line doc strings and [here](https://github.com/stratis-storage/stratis-cli/blob/develop/src/stratis_cli/_errors.py#L74-L75) for multi line doc strings.

tests/whitebox/unittest/test_error_fmt.py
```diff
@@ -0,0 +1,133 @@
...
+
+from stratis_cli._stratisd_constants import BlockDevTiers
+
+
+class ErrorFmtTestCase(unittest.TestCase):
+    """
+    Test stringification of various error types.
+    """
+
+    def StringNotEmpty(self, exception):
```
jbaublitz: This method does not need to be public so I'd name this `_string_not_empty`. Other than that, this method looks good.

tests/whitebox/unittest/test_error_fmt.py
jbaublitz: The type should be something more like [this](https://github.com/stratis-storage/stratis-cli/blob/develop/src/stratis_cli/_error_reporting.py#L88). However, this should be just `Exception` here instead of `list of Exception`. This should be a type instead of a description of what it's used for.

jbaublitz:
@waltdisgrace Let's close this and open a new PR with the commits squashed.

PR #413, opened 2019-11-11
Add isort to Makefile and apply make fmt

jbaublitz:
Supersedes #329 
Related stratis-storage/project#40


jbaublitz:
For now, just rebasing onto develop and seeing if CI passes.

mulkieran:
We can move this up to November. There are some groupings that I don't like, so I'ld like to tinker w/ it before it goes in. OTOH, it's automatic, so if we can accept somethings that we slightly don't like, it'll be the better choice.

mulkieran:
Note that if we choose to do this for stratis-cli, we'll want to do this for all the other Python projects as well.

jbaublitz:
I am slightly confused by some of the reordering. I've been looking at it a bit more to try to figure out exactly how they're being grouped.

mulkieran:
I'm going to test these changes out in one of the smaller libraries, just to get a feel for how it works.

mulkieran:
We can take our time on this, moving to December.

jbaublitz:
Superseded by #422

Issue #303, opened 2019-08-12
Improve our tests of list commands so that they verify something about what is printed


mulkieran:
Right now, they just verify that printing completes w/out raising an exception.

mulkieran:
Actually, much better to write sensible unit tests for our formatting code.

PR #415, opened 2019-11-15
Add table printing tests

waltdisgrace:
This is a preliminary PR that will resolve: #303.


jbaublitz:
@waltdisgrace My suggestion here is that if we're going to use a file-like object, we may want to use an in-memory file-like object instead of worrying about cleaning up actual files on the filesystem if there happens to be a failure. See [`StringIO`](https://docs.python.org/3/library/io.html#text-i-o) as a possible alternative. What are your thoughts, @mulkieran?

mulkieran:
> @waltdisgrace My suggestion here is that if we're going to use a file-like object, we may want to use an in-memory file-like object instead of worrying about cleaning up actual files on the filesystem if there happens to be a failure. See [`StringIO`](https://docs.python.org/3/library/io.html#text-i-o) as a possible alternative. What are your thoughts, @mulkieran?

@jbaublitz You're certainly correct about avoiding actual files, and going with something like StringIO instead. But, this one I'ld like to push back a bit, because I think we want to abstract what we actually test a bit more, and that will require some reflection.

waltdisgrace:
Rebased.

waltdisgrace:
Rebased.

mulkieran:
@waltdisgrace Looks like you're using the wrong version of isort again.

tests/whitebox/unittest/test_formatting.py
```diff
@@ -0,0 +1,91 @@
...
+        print_table(
+            self.table[0], self.table[1:], ["<", "<", "<", "<", "<", "<"], self.output
+        )
+
+    def testContainsEquallyLongRows(self):
+        """
+        Test that the table's rows are of equal length
+        """
+        self.output.seek(0)
+        row_lengths = []
```
mulkieran: I'ld like you to be more functional and less iterative here. We avoid the loop because a loop is a very general structure and much can happen within it. We change these three lines into just one line:

```row_lengths = map(maybe_wcswidth, self.output.readline())```

then it's pretty clear that this is a straightforward application of a function from each element in a list of lines.
Domain is str, range is int.

File: tests/whitebox/unittest/test_formatting.py [hunk 2]
```diff
@@ -0,0 +1,89 @@
...
+            self.table[0], self.table[1:], ["<", "<", "<", "<", "<", "<"], self.output
+        )
+
+    def testContainsEquallyLongRows(self):
+        """
+        Test that the table's rows are of equal length
+        """
+        self.output.seek(0)
+        row_lengths = map(maybe_wcswidth, self.output.readline())
+        self.assertEqual(len(set(row_lengths)), 1)
```
mulkieran: Instead of set, use frozenset. It's more efficient and is used when you don't expect to change the elements in the set after the set is created.

mulkieran:
I have observed the test failures, which are, indeed, a puzzle.

tests/whitebox/unittest/test_formatting.py
```diff
@@ -0,0 +1,89 @@
...
+        print_table(
+            self.table[0], self.table[1:], ["<", "<", "<", "<", "<", "<"], self.output
+        )
+
+    def testContainsEquallyLongRows(self):
+        """
+        Test that the table's rows are of equal length
+        """
+        self.output.seek(0)
+        row_lengths = map(maybe_wcswidth, self.output.readline())
```
mulkieran: There's a typo in this line...it should be ```readlines```. That's probably the source of the problem.

mulkieran:
Travis tests just has a Travis error, so I restarted them.

Issue #419, opened 2019-11-20
Remove all "no cover" pragmas in _errors.py


mulkieran:
We should be able to get complete coverage on every line in this file by extending the unittests in ```test_error_fmt.py```. Currently, we're at 94% on that file, if all remaining "no cover" pragmas are removed.

mulkieran:
The technique here should be just to remove all the "no cover" pragmas and related comments and use the test infrastructure to bring the coverage back up to 100%.

waltdisgrace:
Closed by: #424, #425, #433.

mulkieran:
There's one "no cover" pragma remaining...

waltdisgrace:
Closed by: #424, #425, #433, #437.

PR #421, opened 2019-11-20
Fully cover StratisCliPartialChangeError

waltdisgrace:
Partially addresses #419.

### DIFF
--- a/tests/whitebox/integration/logical/test_create.py
+++ b/tests/whitebox/integration/logical/test_create.py
@@ -114,12 +114,12 @@ class Create3TestCase(SimTestCase):
 
 class Create4TestCase(SimTestCase):
     """
-    Test creating multiple volumes, of which one already exists.
+    Test creating a set of volumes, of which one already exists.
     """
 
     _MENU = ["--propagate", "filesystem", "create"]
     _POOLNAME = "deadpool"
-    _VOLNAMES = ["oubliette", "mnemosyne"]
+    _VOLNAMES = ["oubliette", "mnemosyne", "gaia"]
 
     def setUp(self):
         """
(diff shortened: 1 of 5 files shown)


waltdisgrace:
@mulkieran, with the current edits, Lines 87, 86->87 of _errors.py remain uncovered.

As far as I can tell, the test located in Lines 182-193 of the test_destroy.py and the test located in Lines 183-194 of the test_create.py's file should be covering these lines. What am I not seeing?

jbaublitz:
@mulkieran I have one concern about the filesystem tests, mainly that we'll be passing multiple filesystems into `CreateFilesystems` in our testing framework while stratisd only supports one filesystem per request currently. Do we want to move forward with that approach anyway and then address the single filesystem limit later? I will say that if we do that, the tests will not have to be changed which is a nice benefit, but it may also lead to confusing testing errors in the meantime.

tests/whitebox/integration/logical/test_create.py
```diff
@@ -132,12 +132,76 @@ def setUp(self):
...
+
+    def setUp(self):
+        """
+        Start the stratisd daemon with the simulator.
+        """
+        super().setUp()
+        command_line = ["pool", "create", self._POOLNAME] + _DEVICE_STRATEGY()
+        RUNNER(command_line)
+
+        command_line = self._MENU + [self._POOLNAME] + self._VOLNAMES[0:1]
```
mulkieran: This is the kind of thing that pylint should catch...you assign to ```command_line``` twice in sequence. The first assignment has no effect.

tests/whitebox/integration/logical/test_destroy.py
```diff
@@ -133,10 +133,74 @@ def setUp(self):
...
+
+    def setUp(self):
+        """
+        Start the stratisd daemon with the simulator.
+        """
+        super().setUp()
+        command_line = ["pool", "create", self._POOLNAME] + _DEVICE_STRATEGY()
+        RUNNER(command_line)
+
+        command_line = self._MENU + [self._POOLNAME] + self._VOLNAMES[0:1]
```
mulkieran: Here is the same problem again.

mulkieran:
I think that using multiple filesystem arguments is just fine..., our code is set up to handle that.

tests/whitebox/integration/logical/test_create.py
```diff
@@ -132,12 +132,78 @@ def setUp(self):
...
```
mulkieran: Line 179 and 182 together need a comment to explain why you're calling filesystem create method twice. Otherwise, future readers will be very confused. And, should stratisd allow reation of more than one, then the comment would mark where we could get rid of this odd-looking construction. So it's a good idea  to leave some breadcrumbs for future maintainers here.

tests/whitebox/integration/logical/test_destroy.py
```diff
@@ -133,10 +133,76 @@ def setUp(self):
...
```
mulkieran: The same here as above. Leave a comment to help out future maintainers.

waltdisgrace:
> This looks good! Please go ahead, close this, and open a final PR.

To edit this PR, I've been amending a single, existing commit; do we nevertheless require another PR?

mulkieran:
This PR should have been constructed of multiple commits so the reviewers can easily see the changes as review proceeds. I hadn't noticed that it was a succession of force pushes probably due to the holiday; force-pushing removes the edits that the attached discussion is about, and makes the whole process less clear, and generally slower since the reviewer _has_ to read everything, instead of just the changes they requested.

Please go ahead and open a new, final PR.

Issue #426, opened 2019-11-22
Make a separate subclass of StratisCliRuntimeError for user errors


mulkieran:
Probably call it StratisCliUserError. Make all exceptions that are used to detect a probable error on the user's part subtypes of the new type. Realise that all exceptions that are subtypes of StratisCliUserError can get full coverage without appearing in the unittests.

After this part, work on an additional commit that generates a generic error for all user mistakes. This will require somewhat careful work with inheritance and also modification of methods in ```_error_reporting.py``` so it should be done in a separate step from the first part.

Related: #419.

mulkieran:
@waltdisgrace You are welcome to do the two parts separately. In fact, you might want to begin on the first part immediately, then proceed with the rest of #419, then do the second part.

mulkieran:
Note that you can use the Makefile target, "view" to get a diagram of the exception class hierarchy.

PR #428, opened 2019-11-22
Add StratisCliUserError class

waltdisgrace:
This is a preliminary PR that will address #426.

### DIFF
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -129,17 +129,10 @@ def _interpret_errors(errors):
                 "the D-Bus: %s."
             )
             return fmt_str % error
-        if isinstance(error, StratisCliPartialChangeError):
-            fmt_str = "You issued a command that would have a partial or no effect: %s"
+
+        if isinstance(error, StratisCliUserError):
+            fmt_str = "It appears that you issued an unintended command: %s"
             return fmt_str % error
-        if isinstance(error, StratisCliInUseError):
-            fmt_str = (
-                "You issued a command that would have resulted in "
-                "including a block device in both cache and data tiers: %s"
-            )
-            return fmt_str % error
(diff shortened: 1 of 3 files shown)


waltdisgrace:
@mulkieran, [here](https://github.com/stratis-storage/stratis-cli/files/3887224/classes_stratis-cli-errors.pdf) is a PDF of the class hierarchy that reflects the proposed changes.

src/stratis_cli/_error_reporting.py
```diff
@@ -125,6 +126,10 @@ def _interpret_errors(errors):
                 "the D-Bus: %s."
             )
             return fmt_str % error
+
+        if isinstance(error, StratisCliUserError):
+            fmt_str = "Invalid user input: %s"
+            return fmt_str % error
```
mulkieran: The ```isinstance``` checks for all subclasses of ```StratisCliInUseError``` become unreachable when you check ```isinstance(error, StratisCliInUseError)``` first. Since they can never be reached, you should delete them.

But, you should consider whether some part of the error messages that you see here in, e.g., StratisCliInUseError match block should be made part of the ```StratisCliInUseError```'s ```__str__``` method. You'll want to consider very carefully, so that the error messages will not become less readable.

src/stratis_cli/_error_reporting.py
```diff
@@ -125,6 +126,10 @@ def _interpret_errors(errors):
                 "the D-Bus: %s."
             )
             return fmt_str % error
+
+        if isinstance(error, StratisCliUserError):
+            fmt_str = "Invalid user input: %s"
```
mulkieran: I'm not sure I'ld go with this, because the input _is_ valid, it's the context that makes us decide that it is wrong. I'ld go with something gentler and more explanatory, "It looks like you issued a command that you did not intend", maybe.

waltdisgrace:
Rebased.

PR #435, opened 2019-12-05
Fully cover StratisCliInUseError

waltdisgrace:
Partially addresses #419.

### DIFF
--- a/src/stratis_cli/__init__.py
+++ b/src/stratis_cli/__init__.py
@@ -14,5 +14,5 @@
 """
 Top level of CLI.
 """
-from ._main import run
 from ._errors import StratisCliEnvironmentError
+from ._main import run
--- a/src/stratis_cli/__init__.py
+++ b/src/stratis_cli/__init__.py
@@ -14,5 +14,5 @@
 """
 Top level of CLI.
 """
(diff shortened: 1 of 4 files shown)


waltdisgrace:
> Please leave the `__init__.py` files alone. isort ignores the ordering of imports in them by default; so we are going to do that too, unless we make a global decision to do otherwise. These changes don't have anything to do with the purpose of the PR, so they don't belong here.

I can remove these changes. FYI, these imports were ordered via isort (I just double-checked this by running `make fmt` on the upstream master branch and viewing the diff).

mulkieran:
> > Please leave the `__init__.py` files alone. isort ignores the ordering of imports in them by default; so we are going to do that too, unless we make a global decision to do otherwise. These changes don't have anything to do with the purpose of the PR, so they don't belong here.
> 
> I can remove these changes. FYI, these imports were ordered via isort (I just double-checked this by running `make fmt` on the upstream master branch and viewing the diff).

You should check your isort version. Recall that we're using 4.3.4. Later versions of isort do change their behavior a bit, but we've chosen to go with the version that is the newest in Fedora 30, even though it is not the newest available from PyPi.

Issue #418, opened 2019-11-19
Broken pipe can cause traceback in stratis fs list


drckeefe:
Created a over 2K filesystems on one pool, then ran: stratis fs|head, which cased a traceback. 
```
[root@stratis]# stratis fs|head
Pool Name  Name    Used     Created            Device              UUID
p1         fs382   545 MiB  Nov 19 2019 13:59  /stratis/p1/fs382   35ac31f0393f412ab9ab5f6cf1aad375
p1         fs383   545 MiB  Nov 19 2019 13:59  /stratis/p1/fs383   7754a91bd5e44ac892b875245734f72a
p1         fs384   545 MiB  Nov 19 2019 13:59  /stratis/p1/fs384   a29c5c410bc5451f84f292e56576f79f
p1         fs385   545 MiB  Nov 19 2019 13:59  /stratis/p1/fs385   59bd320515c041df877422fd926ee285
p1         fs386   545 MiB  Nov 19 2019 13:59  /stratis/p1/fs386   08f68080f5d34c8b92a4ba269e9328de
p1         fs387   545 MiB  Nov 19 2019 13:59  /stratis/p1/fs387   6387b1cd3544435988b91044e07fbb16
p1         fs388   545 MiB  Nov 19 2019 13:59  /stratis/p1/fs388   65b26fcbdd3d4d438e8e9af044ce319a
p1         fs389   545 MiB  Nov 19 2019 13:59  /stratis/p1/fs389   779667fa0e9a4faaaeddc0903f065007
p1         fs390   545 MiB  Nov 19 2019 13:59  /stratis/p1/fs390   ea1e8320836446c3be0fa99077e3fc94
stratis encountered an unexpected error during execution. Please report the error and include in your report the stack trace shown below.

Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/stratis_cli/_main.py", line 43, in the_func
    result.func(result)
  File "/usr/lib/python3.6/site-packages/stratis_cli/_actions/_logical.py", line 178, in list_volumes
    ["<", "<", "<", "<", "<", "<"],
  File "/usr/lib/python3.6/site-packages/stratis_cli/_actions/_formatting.py", line 164, in print_table
    _print_row(file, row, row_widths, column_widths, alignment)
  File "/usr/lib/python3.6/site-packages/stratis_cli/_actions/_formatting.py", line 126, in _print_row
    print("  ".join(entries), end="", file=file)
[...]

PR #431, opened 2019-11-27
Handle BrokenPipeError

waltdisgrace:
Resolves #418.

Demonstration of handled error:
[root@vdo-storage-02 stratis_cli]# stratis fs | head
Pool Name  Name    Used     Created            Device              UUID                            
p1         fs928   545 MiB  Nov 26 2019 10:11  /stratis/p1/fs928   e03b0893d11a468e8d0f59f433e0758b
p1         fs1206  545 MiB  Nov 26 2019 10:21  /stratis/p1/fs1206  fda1c158e3414edd93d600dda778068c
p1         fs444   545 MiB  Nov 26 2019 09:58  /stratis/p1/fs444   00ce78c7f7b1460d8758ed44af39590b
p1         fs466   545 MiB  Nov 26 2019 09:59  /stratis/p1/fs466   2c6f19978434405e9b9961d69ccfc8c4
p1         fs1093  545 MiB  Nov 26 2019 10:17  /stratis/p1/fs1093  30a90088a6564ba681fd8ddb7bd1f1ca
p1         fs1205  545 MiB  Nov 26 2019 10:21  /stratis/p1/fs1205  1444f157ab1a4e999c18ceb81854bfcc
p1         fs1216  545 MiB  Nov 26 2019 10:21  /stratis/p1/fs1216  5cd8e208901b42abbe23758e85ae59d5
p1         fs438   545 MiB  Nov 26 2019 09:58  /stratis/p1/fs438   4dc195c15c1f43daaf07d971d69470c1
p1         fs323   545 MiB  Nov 26 2019 09:56  /stratis/p1/fs323   e8d60d212d92427f872bfa570aaeabf7
Execution failed:
Broken pipe

PR #440, opened 2019-12-09
Handle BrokenPipeError

waltdisgrace:
Supersedes: #431.
This is a preliminary PR that will resolve: #418.


mulkieran:
I found some more useful information, so I think we'll have to think this through one more time: https://docs.python.org/3.8/library/signal.html

mulkieran:
Ok! We will have to revise one more time.

Issue #451, opened 2020-01-07
Document the meaning of "FAILURE" occurring in any field in the man pages


mulkieran:
We already document "???", but "FAILURE" means something different, i.e., "the value was unavailable" and not "the value was not understood".

PR #452, opened 2020-01-08
Update man page: Add description of FAILURE unknown value

bgurney-rh:
Resolves: #451

### DIFF
--- a/docs/stratis.txt
+++ b/docs/stratis.txt
@@ -149,10 +149,15 @@ Mailing list::
 	stratis-devel@lists.fedorahosted.org for general development
 	discussion
 Unknown values::
-	stratis-cli uses the Stratis API. If the API returns values that
-	stratis-cli cannot interpret, stratis-cli will substitute "???". If
-	encountered, upgrading to the latest version of stratis-cli, or filing
-	an issue, is recommended.
+	If the stratisd D-Bus API returns values that stratis-cli cannot
+	interpret, stratis-cli will substitute "???". If encountered,
+	upgrading to the latest version of stratis-cli, or filing an
+	issue, is recommended.
+Unobtainable values::
+	If the stratisd D-Bus API indicates that a value is unobtainable,
+	stratis-cli will substitute "FAILURE". This may indicate something
+	wrong with the pool, blockdev, or filesystem. In some cases,
+	restarting stratisd may resolve the issue.
 
 LICENSE
 -------


docs/stratis.txt
```diff
@@ -151,8 +151,9 @@ Mailing list::
 Unknown values::
 	stratis-cli uses the Stratis API. If the API returns values that
 	stratis-cli cannot interpret, stratis-cli will substitute "???". If
-	encountered, upgrading to the latest version of stratis-cli, or filing
-	an issue, is recommended.
+	the API detects that a value is unobtainable, stratis-cli will
```
mulkieran: I think that "???" and "FAILURE" mean such different things that the recommendation should be different. For "???" update the CLI and then, if that doens't fix the problem, file an issue, is a good suggestion.

But for "FAILURE", if the value is unobtainable, then something is likely wrong with the pool or blockdev or filesystem. So, the user should be warned somehow, and possibly strongly encouraged to file an issue explaining how they got to the point where list commands had "FAILURE" entries.

So you want to separate out these two different non-values more strongly, at least with separate paragraphs, if not w/ separate headings.

bgurney-rh: Okay, I can split them.  (Though, both seem to fall under the "stratis-cli uses the Stratis API" preamble, but I don't know if it'll be too redundant to state that twice.)

How does this look, just after the original "Unknown Values" heading?
```
Unobtainable values::
        If the API detects that a value is unobtainable, stratis-cli will
        substitute "FAILURE". This may indicate something wrong with the pool,
        blockdev, or filesystem.
```

mulkieran: I think that's good. But you made me realize that unknown values preamble is a bit off, too.

Instead of

```
stratis-cli uses the Stratis API. If the API returns values that
	stratis-cli cannot interpret...
```

Try just:

```
If the stratisd D-Bus API returns a value that stratis-cli cannot interpret,...
```

and then for failure
```
If the stratisd D-Bus API returns a message indicating that some value could not be obtained...
```

or something like that.

bgurney-rh: OK; I'm updating it now.

docs/stratis.txt
```diff
@@ -149,10 +149,14 @@ Mailing list::
...
-	stratis-cli uses the Stratis API. If the API returns values that
-	stratis-cli cannot interpret, stratis-cli will substitute "???". If
-	encountered, upgrading to the latest version of stratis-cli, or filing
-	an issue, is recommended.
+	If the stratisd D-Bus API returns values that stratis-cli cannot
+	interpret, stratis-cli will substitute "???". If encountered,
+	upgrading to the latest version of stratis-cli, or filing an
+	issue, is recommended.
+Unobtainable values::
+	If the stratisd D-Bus API detects that a value is unobtainable,
```
mulkieran: ^detects^indicates..the D-Bus API layer is really just passing information through from the engine, so it should not really detect anything.

bgurney-rh: Ah, okay; updated.

mulkieran:
> Given that we've separated the two out, do we want to suggest a remedy step like restarting stratisd? That obviously will not always work, but it may resolve a good percentage of potential issues.

Yes, that seems like a good idea.

bgurney-rh:
I added the following: "In some cases, restarting stratisd may resolve the issue."

jbaublitz:
That sounds good to me.

bgurney-rh:
The previous Jenkins test failure was due to the test system having "rustup default 1.31.0" installed.  After updating to the current minimum rust version (1.38.0), the test passes.

PR #458, opened 2020-01-27
Abstract error code check

lleshchi:
Related stratis-storage/stratis-cli#371

Issue #371, opened 2019-10-23
Check the error code from invocation of RUNNER in dbus tests


mulkieran:
We can and should do this. We rely on the implicit test of no exception raised, but checking the error code as well would be better.

waltdisgrace:
RUNNER is equal to the the_func method defined [here](https://github.com/stratis-storage/stratis-cli/blob/f77c5fcde159b20c1ef3f084727fc818dffeb4a0/src/stratis_cli/_main.py#L36), which only ever raises an error or returns 0. This, coupled with the fact that we rely on the implicit test that no exception is raised, seems to imply that checking the error code would be redundant.

What is the reason that checking the error code would be better?

mulkieran:
This is a bit of a tricky area, because of the fluid boundaries between the Python interpreter and the program. Right now our program only returns 0 (good), 1(error), 2(parser error), and it is correct that this method can only return a 0. All other codes returned by the program are handled by the interpreter through the raising of a SystemExit exception of some sort.

So, what I'm really saying here, is that we should somehow be checking the error code associated with the SystemExit exception. But, since right now it will always be either 2 or 1, so there is no great hurry. We actually do test it in a few cases...probably we should just be testing it in more.

mulkieran:
For most of our tests, ```handle_err``` in the ```run()``` method is never reached, because we choose to use the ```--propagate``` flag for most tests. The reason for this is that we get more benefit from our tests this way, because we can check the resulting exception precisely. However, in most cases, we could enhance our existing tests by taking the caught exception and just passing it to the ```handle_err``` method.

Something like:

```
with self.assertRaises(SystmeExit) as final_err:
    handle_err(err)    # this is the error that we caught previously in the tests and it ought always to have been a StratisCliActionError
# assert that the exit code associated with final_err is 1.
# if it isn't, or we don't catch SystemExit then there is some kind of bug
````

You can find _some_ examples in the code where I deal with SystemExit exceptions, but I expect almost exclusively to verify that parsing errors are being handled correctly.

mulkieran:
Related enhancement request: https://bugs.python.org/issue36691.

mulkieran:
Probably the definition of the error codes should be in the source rather than in the tests, since we intend to extend the values in future to cover more than 0, 1, and 2.

PR #459, opened 2020-01-29
Check error code handling

lleshchi:
Created an exit_() function to replace the standard SystemExit exception. This function will accept both a exit code and an error message, which will be printed to stderr. 

The exit_() function is now used within hande_error() and has been tested with the check_handle_error() function within one of the test_create.py unit tests. This check needs to be extended to other unit tests. 

Additionally, all other calls to sys.exit() should be replaced with exit_(). 

Related stratis-storage/stratis-cli#371 
Supersedes stratis-storage/stratis-cli#458

### DIFF
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -58,9 +58,8 @@ _DBUS_INTERFACE_MSG = (
 
 def exit_(code, msg):
     """
-    Description
+    Exits program with a given exit code and error message.
     """
-    # if msg is not None
     print(msg, file=sys.stderr)
     raise SystemExit(code)
 
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -48,6 +48,8 @@ class StratisCliErrorCodes(IntEnum):
     PARSE_ERROR = 2
 
 
+ERROR = StratisCliErrorCodes.ERROR
+
 _DBUS_INTERFACE_MSG = (
     "The version of stratis you are running expects a different "
     "D-Bus interface than the one stratisd provides. Most likely "
@@ -60,7 +62,7 @@ def exit_(code, msg):
     """
     Exits program with a given exit code and error message.
     """
-    print(msg, file=sys.stderr)
+    print(msg, os.linesep, file=sys.stderr, flush=True)
     raise SystemExit(code)
 
 
@@ -234,4 +236,4 @@ def handle_error(err):
         raise err
 
     exit_msg = "Execution failed:%s%s" % (os.linesep, explanation)
-    exit_(1, exit_msg)
+    exit_(ERROR, exit_msg)
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -62,7 +62,7 @@ def exit_(code, msg):
     """
     # if msg is not None
     print(msg, file=sys.stderr)
-    raise SystemExit(1)
+    raise SystemExit(code)
 
 
 # pylint: disable=fixme
@@ -235,4 +235,4 @@ def handle_error(err):
         raise err
 
     exit_msg = "Execution failed:%s%s" % (os.linesep, explanation)
-    sys.exit(exit_msg)
+    exit_(1, exit_msg)
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -73,6 +73,9 @@ def _interface_name_to_common_name(interface_name):
     :returns: a common name
     :rtype: str
     """
+    # pylint: disable=fixme
+    # FIXME: remove no coverage pragma when adequate testing for CLI output
+    # exists.
     if interface_name == BLOCKDEV_INTERFACE:  # pragma: no cover
         return "block device"
 
@@ -165,6 +168,9 @@ def _interpret_errors(errors):
         # Inspect lowest error
         error = errors[-1]
 
+        # pylint: disable=fixme
+        # FIXME: remove no coverage pragma when adequate testing for CLI output
+        # exists.
         if (
             # pylint: disable=bad-continuation
             isinstance(error, dbus.exceptions.DBusException)
@@ -181,6 +187,9 @@ def _interpret_errors(errors):
         ):
             return "Most likely stratis is unable to connect to the stratisd D-Bus service."
 
+        # pylint: disable=fixme
+        # FIXME: remove no coverage pragma when adequate testing for CLI output
+        # exists.
         if (
             # pylint: disable=bad-continuation
             isinstance(error, dbus.exceptions.DBusException)
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -73,7 +73,7 @@ def _interface_name_to_common_name(interface_name):
     :returns: a common name
     :rtype: str
     """
-    if interface_name == BLOCKDEV_INTERFACE:
+    if interface_name == BLOCKDEV_INTERFACE:  # pragma: no cover
         return "block device"
 
     if interface_name == FILESYSTEM_INTERFACE:
@@ -82,7 +82,7 @@ def _interface_name_to_common_name(interface_name):
     if interface_name == POOL_INTERFACE:
         return "pool"
 
-    raise StratisCliUnknownInterfaceError(interface_name)
+    raise StratisCliUnknownInterfaceError(interface_name)  # pragma: no cover
 
 
 def get_errors(exc):
@@ -168,7 +168,7 @@ def _interpret_errors(errors):
             # pylint: disable=bad-continuation
             isinstance(error, dbus.exceptions.DBusException)
             and error.get_dbus_name() == "org.freedesktop.DBus.Error.AccessDenied"
-        ):
+        ):  # pragma: no cover
             return "Most likely stratis has insufficient permissions for the action requested."
         # We have observed two causes of this problem. The first is that
         # stratisd is not running at all. The second is that stratisd has not
@@ -184,7 +184,7 @@ def _interpret_errors(errors):
             # pylint: disable=bad-continuation
             isinstance(error, dbus.exceptions.DBusException)
             and error.get_dbus_name() == "org.freedesktop.DBus.Error.NoReply"
-        ):
+        ):  # pragma: no cover
             fmt_str = (
                 "stratis attempted communication with the daemon, stratisd, "
                 "over the D-Bus, but stratisd did not respond in the allowed time."
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -17,6 +17,7 @@ Facilities for managing and reporting errors.
 # isort: STDLIB
 import os
 import sys
+from enum import IntEnum
 
 # isort: THIRDPARTY
 import dbus
@@ -35,6 +36,17 @@ from ._errors import (
     StratisCliUnknownInterfaceError,
     StratisCliUserError,
 )
+
+
+class StratisCliErrorCodes(IntEnum):
+    """
+    StratisCli Error Codes
+    """
+
+    OK = 0
+    ERROR = 1
+    PARSE_ERROR = 2
+
 
 _DBUS_INTERFACE_MSG = (
     "The version of stratis you are running expects a different "
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -64,9 +64,7 @@ def exit_(code, msg):
     raise SystemExit(code)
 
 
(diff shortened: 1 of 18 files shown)


src/stratis_cli/_error_reporting.py
```diff
@@ -56,6 +56,15 @@ class StratisCliErrorCodes(IntEnum):
 )
 
 
+def exit_(code, msg):
+    """
+    Description
+    """
+    # if msg is not None
+    print(msg, file=sys.stderr)
```
mulkieran: Take a look at the other place where we explicitly print. That invocation is very explicit with a bunch of parameters; it's probably best to be that way here, too.

src/stratis_cli/_error_reporting.py
```diff
@@ -56,6 +56,15 @@ class StratisCliErrorCodes(IntEnum):
 )
 
 
+def exit_(code, msg):
```
mulkieran: Why ```exit_```, the underscore after the name? Is there a convention that this is following?

src/stratis_cli/_error_reporting.py
```diff
@@ -214,4 +234,4 @@ def handle_error(err):
         raise err
 
     exit_msg = "Execution failed:%s%s" % (os.linesep, explanation)
-    sys.exit(exit_msg)
+    exit_(1, exit_msg)
```
mulkieran: Use ```StratisCliErrorCodes``` value here.

tests/whitebox/integration/logical/test_create.py
```diff
@@ -32,6 +33,19 @@
 from .._misc import RUNNER, SimTestCase, device_name_list
 
 _DEVICE_STRATEGY = device_name_list(1)
+ERROR = StratisCliErrorCodes.ERROR
+
+
+def check_handle_error(obj, context, expected_code):
```
mulkieran: I'll come and talk about this one, the topic is a bit complicated.

lleshchi:
The following are special cases that don't exactly fit the check_error() format, due to the final line where the return code of the cause is checked. This could perhaps be incorporated into check_error() as an optional parameter: 

https://github.com/stratis-storage/stratis-cli/blob/99abe4478d7fa7e9db1b7de6863a4fd1bdb92812/tests/whitebox/integration/logical/test_create.py#L109-L119

https://github.com/stratis-storage/stratis-cli/blob/99abe4478d7fa7e9db1b7de6863a4fd1bdb92812/tests/whitebox/integration/pool/test_destroy.py#L99-L108

tests/whitebox/_misc.py
```diff
@@ -37,6 +37,32 @@
...
+    :param expected_cause: the expected cause of the top level exception
+    :type expected_cause: StratisCliError
+    :param command_line: the command line arguments
+    :param expected_code: the expected error code
+    :type expected_code: int
+    """
+    with obj.assertRaises(expected_err) as context:
+        RUNNER(command_line)
+
+    cause = context.exception.__cause__
```
mulkieran: You use ```context.exception``` expression twice here; try to use it just once.

tests/whitebox/_misc.py
```diff
@@ -37,6 +37,32 @@
     sys.exit(message)
 
 
+def check_error(obj, expected_err, expected_cause, command_line, expected_code):
+    """
+    Check that the expected exception was raised, and that the cause
+    and exit codes where also as expected, based on the command line arguments
+    passed to the program.
+    :param expected_err: the expected top level exception in the chain
+    :type expecter_err: StratisCliError
```
mulkieran: I think this type can be precisely ```StratisCliActionError```.

tests/whitebox/_misc.py
```diff
@@ -37,6 +37,32 @@
...
+    :param expected_cause: the expected cause of the top level exception
+    :type expected_cause: StratisCliError
```
mulkieran: This type is actually more general, go with ```Exception```, although that might be too general.

tests/whitebox/_misc.py
```diff
@@ -37,6 +37,32 @@
...
+    :param command_line: the command line arguments
```
mulkieran: You can add the type of the command line arguments, which is ```list```.

tests/whitebox/integration/logical/test_rename.py
```diff
@@ -19,11 +19,13 @@
...
 
 # isort: LOCAL
+from stratis_cli import StratisCliErrorCodes
 from stratis_cli._errors import StratisCliActionError, StratisCliNoChangeError
 
-from .._misc import RUNNER, SimTestCase, device_name_list
+from .._misc import RUNNER, SimTestCase, check_error, device_name_list
 
 _DEVICE_STRATEGY = device_name_list(1)
+ERROR = StratisCliErrorCodes.ERROR
```
mulkieran: Please prepend underscores for your ```ERROR``` constants in every file, like ```_ERROR```. This indicates that they are intended to be constants that are used only locally, in the file in which they are defned.

tests/whitebox/_misc.py
lleshchi: And this is to account for DbusClientError as well, correct? Is it bad practice to be too general? Is there a better practice if you only want to accept DbusClientError and StratisCliError?

src/stratis_cli/_error_reporting.py
```diff
@@ -94,7 +114,7 @@ def _interpret_errors(errors):
     # pylint: disable=fixme
     # FIXME: remove no coverage pragma when adequate testing for CLI output
     # exists.
-    try:  # pragma: no cover
+    try:
```
mulkieran: You can remove the three comment lines above here, which are all an explanation for the removed "no cover" pragma.

mulkieran: Oh, and you can remove the "no cover" pragma on the ```_interface_name...``` method as well.

mulkieran:
There will be some things that are still not covered in ```_error_reporting.py```. Our plan will be to add ```no cover``` pragmas on each of these individual code chunks, and then to extend our testing so that they are actually covered. But that is outside of the scope of this PR.

mulkieran:
I think you can drop the following tests. Please do this in a separate commit. These are all tests that are in ```test_error_fmt.py```.

The test for:
* StratisCliActionError
* StratisCliEngineError
* StratisCliRuntimeError
* StratisCliError

Wait to do this, until you're back to 100% code coverage, though, so you can easily detect if dropping these tests causes coverage to be dropped somewhere.

mulkieran:
@lleshchi : Since you're dealing a lot with error classes, it might be nice to get a visual on the error hierarchy. There is a make target called "view" that should help you with that. Look in the directory ```_pyreverse``` for the output.

src/stratis_cli/_error_reporting.py
```diff
@@ -55,7 +73,7 @@ def _interface_name_to_common_name(interface_name):  # pragma: no cover
     :returns: a common name
     :rtype: str
     """
-    if interface_name == BLOCKDEV_INTERFACE:
+    if interface_name == BLOCKDEV_INTERFACE:  # pragma: no cover
```
mulkieran: Put a FIXME here, to mark it as a definite "should cover".

src/stratis_cli/_error_reporting.py
```diff
@@ -153,7 +169,7 @@ def _interpret_errors(errors):
             # pylint: disable=bad-continuation
             isinstance(error, dbus.exceptions.DBusException)
             and error.get_dbus_name() == "org.freedesktop.DBus.Error.AccessDenied"
-        ):
+        ):  # pragma: no cover
```
mulkieran: This should get  a FIXME, too. We can get this easily by running without adequate permissions.

src/stratis_cli/_error_reporting.py
```diff
@@ -169,7 +185,7 @@ def _interpret_errors(errors):
             # pylint: disable=bad-continuation
             isinstance(error, dbus.exceptions.DBusException)
             and error.get_dbus_name() == "org.freedesktop.DBus.Error.NoReply"
-        ):
+        ):  # pragma: no cover
```
mulkieran: Put a FIXME here, too. We could probably test this by setting the timeout environment variable extremely low.

src/stratis_cli/_errors.py
```diff
@@ -330,7 +330,7 @@ def __init__(self, command_line_args, namespace):
         self.command_line_args = command_line_args
         self.namespace = namespace
 
-    def __str__(self):
+    def __str__(self):  # pragma: no cover
```
mulkieran: Put a FIXME here, as well, although this one might be more of a puzzle to test.

tests/whitebox/_misc.py
```diff
@@ -37,6 +38,32 @@
     sys.exit(message)
 
 
+def check_error(obj, expected_cause, command_line, expected_code):
+    """
+    Check that the expected exception was raised, and that the cause
+    and exit codes where also as expected, based on the command line arguments
+    passed to the program.
+    :param expected_cause: the expected cause of the top level exception
```
mulkieran: This is really the exception below the StratisCliActionError exception.

tests/whitebox/_misc.py
```diff
@@ -37,6 +38,32 @@
     sys.exit(message)
 
 
+def check_error(obj, expected_cause, command_line, expected_code):
```
mulkieran: Here, you should document the ```obj``` parameter as well, since it is not ```self```. It's type is ```unittest.TestCase```.

tests/whitebox/integration/test_parser.py
```diff
@@ -18,8 +18,29 @@
...
 
+# isort: LOCAL
+from stratis_cli._error_reporting import StratisCliErrorCodes
+
 from ._misc import RUNNER, SimTestCase
 
+_PARSE_ERROR = StratisCliErrorCodes.PARSE_ERROR
+
+
+def check_parse_error(obj, command_line, expected_code):
```
mulkieran: You should document ```obj``` here as well.

tests/whitebox/integration/test_parser.py
```diff
@@ -18,8 +18,29 @@
...
+    parser error.
+
+    :param command_line: the command line arguments
+    :param expected_code: the expected exit code
+    :type expected_code: int
+    """
+    with obj.assertRaises(SystemExit) as context:
+        RUNNER(command_line)
+    exit_code = context.exception.code
+    obj.assertEqual(exit_code, expected_code)
```
mulkieran: You should hard-code the ```expected_code``` as ```PARSE_ERROR```, I think. In this case, the method is crafted just for these exceptions, handled in this way.

tests/whitebox/_misc.py
lleshchi: Can't we be more specific and have the type be SimTestCase? (I believe this question won't be relevant once we implement the new intermediary class)

mulkieran: Since we're planning to make it an instance method of ```RunTestCase``` anyway, then its type will be ```RunTestCase```, so just ignore this comment.

tests/whitebox/integration/test_parser.py
mulkieran: You can ignore this comment as well.

tests/whitebox/_misc.py
lleshchi: But just for my better understanding, if we were not implementing `RunTestCase`, but rather keeping `check_error` as an external function, it would be correct to have the `obj` type as `SimTestCase`?

mulkieran: We would still use the more general type. There is nothing obviously inherent in this method that makes it only suitable for tests that would require the daemon to be running. You have one counter-example: look at ```testPropagate``` in PropagateTestCase.

lleshchi: Understood, thank you!

mulkieran:
Other methods that would be all the better for being given similar treatment:

* test_stratis
  * testErrorOnMissingFilesystem
  * testNotPropagate

They'll have to be adapted somewhat, though.

lleshchi:
Additionally, `check_parse_error()` needs a new name

tests/whitebox/_misc.py
```diff
@@ -37,6 +38,34 @@
...
```
mulkieran: Whoops! You need to remove this method now.

mulkieran:
> Additionally, `check_parse_error()` needs a new name

Call it ```check_system_exit```?

src/stratis_cli/_errors.py
```diff
@@ -330,7 +330,9 @@ def __init__(self, command_line_args, namespace):
         self.command_line_args = command_line_args
         self.namespace = namespace
 
-    def __str__(self):
+    # pylint: disable=fixme
```
mulkieran: Also, explain what needs to be fixed here...that coverage should be obtained, you should use the same formula as the others.

PR #460, opened 2020-01-29
No longer use tox for CI

mulkieran:
Succeeds: #455.

The previous PR did not remove the dependency on tox, but this one does.


mulkieran:
one more fix...

PR #464, opened 2020-02-06
Generalize exit

lleshchi:
For instances in which a SystemExit exception is raised,
it is useful to introduce a function that can be passed the error code
with which to exit the program, as well as the error message that is
printed to standard error. The SystemExit exception itself can only
accept either an exit code or a exit message, but not both, hence the
introduction of a new function.

Additionally, it is beneficial to have the error codes of the program
accessible as constants, especially if additional error codes are introduced
in the future.

Signed-off-by: Leah Leshchinsky <lleshchi@redhat.com>

### DIFF
--- a/bin/stratis
+++ b/bin/stratis
@@ -23,6 +23,7 @@ try:
 
     from stratis_cli import run
     from stratis_cli import StratisCliEnvironmentError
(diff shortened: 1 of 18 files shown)


mulkieran:
The only cause of failure on Jenkins is the coverage failure, so that looks good.

Issue #392, opened 2019-11-01
I'ld prefer to use psutil in blackbox tests


mulkieran:
It's a well-established Python package, it beats hand-coding this ```process_exists``` method that is used instead. But this is hardly the top priority.

PR #471, opened 2020-02-11
Implement psutil within process_exists function

lleshchi:
Related stratis-storage/stratis-cli#392
Needs to be tested by Bryan

Signed-off-by: Leah Leshchinsky <lleshchi@redhat.com>

### DIFF
--- a/tests/blackbox/testlib/utils.py
+++ b/tests/blackbox/testlib/utils.py
@@ -59,7 +59,7 @@ def process_exists(name):
     Look through processes, using their pids, to find one matching 'name'.
     Return None if no such process found, else return the pid.
     :param name: name of process to check
-    :type name: string
+    :type name: str
     :return: pid or None
     :rtype: int or NoneType
     """
--- a/tests/blackbox/testlib/utils.py
+++ b/tests/blackbox/testlib/utils.py
@@ -60,7 +60,8 @@ def process_exists(name):
     Return None if no such process found, else return the pid.
     :param name: name of process to check
     :type name: string
-    :return: None or pid
+    :return: pid or None
+    :rtype: int or NoneType
     """
     for pid in psutil.pids():
         try:
--- a/tests/blackbox/testlib/utils.py
+++ b/tests/blackbox/testlib/utils.py
@@ -63,7 +63,7 @@ def process_exists(name):
     :return: pid or None
     :rtype: int or NoneType
     """
-    for proc in psutil.process_iter():
+    for proc in psutil.process_iter(["name"]):
         try:
             if proc.name() == name:
                 return proc.pid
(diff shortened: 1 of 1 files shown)


tests/blackbox/testlib/utils.py
```diff
@@ -53,16 +56,19 @@ def random_string(length=4):
 
 def process_exists(name):
     """
-    Walk the process table looking for executable 'name', returns pid if one
-    found, else return None
+    Look through processes, using their pids, to find one matching 'name'.
+    Return None if no such process found, else return the pid.
+    :param name: name of process to check
+    :type name: string
+    :return: None or pid
```
mulkieran: You can also use the ```rtype``` markup, which should say something like: ```NoneType or int```

lleshchi: Is this in addition to `return`, or instead of it?

mulkieran: In addition: ```:param``` is to ```:type``` as ```:return``` is to ```:rtype".

lleshchi:
May `process_iter` be extended to the whitebox tests as well?

bgurney-rh:
I was able to successfully test this; however, I had to manually install python3-psutil to get it working.

I created a CI PR to have this package installed by the stratis-blackbox-run.sh script.

bgurney-rh:
The CI PR that installs python3-psutil as part of running stratis-blackbox-run.sh has been merged, so that part should be all set.

tests/blackbox/testlib/utils.py
```diff
@@ -53,16 +56,20 @@ def random_string(length=4):
 
 def process_exists(name):
     """
-    Walk the process table looking for executable 'name', returns pid if one
-    found, else return None
+    Look through processes, using their pids, to find one matching 'name'.
+    Return None if no such process found, else return the pid.
+    :param name: name of process to check
+    :type name: string
```
jbaublitz: @mulkieran do we want this to be `str`?

mulkieran: > @mulkieran do we want this to be `str`?

Yes, it should be ```str```.

tests/blackbox/testlib/utils.py
```diff
@@ -53,16 +56,20 @@ def random_string(length=4):
...
+    :return: pid or None
+    :rtype: int or NoneType
     """
-    for pid in [pid for pid in os.listdir("/proc") if pid.isdigit()]:
+    for proc in psutil.process_iter():
```
mulkieran: If you do use ```process_iter``` you might as well get the benefit. I think there is some design where you can specify that it should just get the name, and not all the information about the process, which is far more efficient. You might as well try that out.

mulkieran:
> May `process_iter` be extended to the whitebox tests as well?

That would be fine, but we'ld file an issue and do it in a separate PR, just because the blackbox tests are so special.

PR #473, opened 2020-02-11
Update documentation to match current blackbox tests

lleshchi:


tests/blackbox/README.md
```diff
@@ -1,11 +1,17 @@
...
+
+The `stratisd` test may be run without the installation of the `stratis` command line.
+
+```bash
+# python3 stratisd_cert.py -v --disk /dev/vdb --disk /dev/vdc --disk /dev/vdd
+```
+
+The stratis-cli test assumes that the `stratisd` daemon is installed, although the test will start the daemon if needed.
 
 Run the test as root and supply 3 or more block devices on the
```
mulkieran: These lines are actually true for both, they must both be root and must be supplied with three or more block devices. Try   putting this line at the top or the bottom, so that it does cover both.

PR #474, opened 2020-02-14
Omit ErrorHandlingTestCase

mulkieran:
Succeeds #465


jbaublitz:
What was the rationale for removing this? It looks like it got put into motion while I was away.

mulkieran:
> What was the rationale for removing this? It looks like it got put into motion while I was away.

#465 introduced the ```check_error``` method, which does what this test method does for all tests where it is called. The test I'm removing here was, in a sense, a prototype of the ```check_error``` method, used for a specific case.

mulkieran:
I'm going to pull this back, and add a precondition to the ```check_error``` method, for clarity.

Issue #470, opened 2020-02-08
Have tests cover all code that should reasonably be covered


mulkieran:
- [x] #469 
- [x] #468 
- [x] #466

mulkieran:
Closing, since there is only one sub-issue left.

PR #475, opened 2020-02-17
Expand test coverage

### LINKED ISSUE #468: Cover _interface_name_to_common_name uncovered branch by writing a test that specifies a non-existant block device
I believe that this should be quite straightforward.

### LINKED ISSUE #469: Write a test that invokes the __str__ method  of the StratisCliActionError, which is the head of the exception chain
Our tests seem to be built to never do that, since this error just contains context about how the method was called. This one is potentially a bit tricky, and might involve manipulating the behavior of the interpreter. But it might turn out to involve something pretty straightforward as well.

lleshchi:
Resolves #468
Resolves #469 
Related #470

### DIFF
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -73,9 +73,10 @@ def _interface_name_to_common_name(interface_name):
     :returns: a common name
     :rtype: str
     """
-    # pylint: disable=fixme
-    # FIXME: remove no coverage pragma when adequate testing for CLI output
-    # exists.
+    # There is no action which specifies a block device once it has been
+    # added to a pool. Consequently, we cannot test that an non-existant block
+    # device has been specified. In the future, if such an action is created,
+    # this no cover may be removed.
     if interface_name == BLOCKDEV_INTERFACE:  # pragma: no cover
         return "block device"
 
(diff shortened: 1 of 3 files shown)


tests/whitebox/_misc.py
```diff
@@ -112,6 +112,10 @@ def check_error(self, expected_cause, command_line, expected_code):
         cause = exception.__cause__
         self.assertIsInstance(cause, expected_cause)
 
+        error_string = str(exception)
+        self.assertNotEqual(error_string, None)
```
mulkieran: ```self.assertNotEqual(error_string, "")``` here. ```None``` has type ```NoneType```, so if ```error_string``` is an instance of a ```str``` it will certainly not equal ```None```.

src/stratis_cli/_error_reporting.py
```diff
@@ -73,9 +73,10 @@ def _interface_name_to_common_name(interface_name):
     :returns: a common name
     :rtype: str
     """
-    # pylint: disable=fixme
-    # FIXME: remove no coverage pragma when adequate testing for CLI output
-    # exists.
+    # There is no action which specifies a block device once it has been
+    # added to a pool. Consequently, we cannot test that an non-existant block
```
mulkieran: typo: ^an non-existant^a non-existent

PR #414, opened 2019-11-15
Initial encryption work

jbaublitz:
Initial review only.

PR #486, opened 2020-02-24
Initial encryption work

jbaublitz:
Supersedes #414 

Blocked on stratisd encryption work in stratis-storage/stratisd#1817

### DIFF
--- a/docs/stratis.txt
+++ b/docs/stratis.txt
@@ -31,7 +31,7 @@ GLOBAL OPTIONS
 
 COMMANDS
 --------
-pool create [--redundancy <redundancy>] <pool_name> <blockdev> [<blockdev>..]::
+pool create [--redundancy <redundancy>] [--key-desc <key_desc>] <pool_name> <blockdev> [<blockdev>..]::
      Create a pool from one or more block devices, with the given pool name.
 pool list::
      List all pools on the system.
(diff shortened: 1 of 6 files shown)


jbaublitz:
@bgurney-rh The CI should (and will) fail.

jbaublitz:
@mulkieran I'm going to go ahead and update the parameter to `--key-desc` to save some time after stratis-storage/stratisd#1908 is merged.

jbaublitz:
> Please touch up the man pages as well while you're here.

What are you looking for in terms of changes? It doesn't look like we document flags in the man pages so are you looking for a longer form description of encryption?

mulkieran:
> > Please touch up the man pages as well while you're here.
> 
> What are you looking for in terms of changes? It doesn't look like we document flags in the man pages so are you looking for a longer form description of encryption?

```--key-desc``` is the only meaningful option that we have had in a long time. We have one meaningless option, "--redundancy", hanging around, though, and it is so meaningless that we haven't noticed that we don't document it. I think it's better to document ```--redundancy``` rather than drop it; so I'll do that and that should provide a template for documentation of options.

mulkieran:
Also necessary to add the init_cache command line argument.

jbaublitz:
@mulkieran Both the command and the man page should be ready for a glance.

jbaublitz:
Coverage should be fixed and I also found a bug that should be resolved in #501

mulkieran:
> Whoops! I missed a spot. Can you also do the obvious alteration to the blackbox tests. StratisDbus.create_pool should get a new parameter.

jbaublitz: Actually, please ignore this request. We can handle the stratisd tests in a separate PR, which will redirect to the new interfaces, I think. But it is possible that the CLI tests should be looked at.

mulkieran:
@jbaublitz: I think, but am not yet certain, that the only change necessary to keep the blackbox tests from failing is to modify ```StratisCertify.test_pool_add_cache``` appropriately, by changing it so that it expects a failure.

jbaublitz:
Okay, will do!

PR #489, opened 2020-03-02
Display some properties of the pool using an encoding

mulkieran:
The CLI display looks like this:

```
# PYTHONPATH=./src ./bin/stratis pool list
Name                  Total Physical   Properties
name         512 EiB / 0 B / 512 EiB   ~Ca,~Cr   
with_cache   512 EiB / 0 B / 512 EiB   Ca,~Cr    
```

Ca indicates that the pool has a cache, Cr indicates that the pool is encrypted.

### DIFF
--- a/docs/stratis.txt
+++ b/docs/stratis.txt
@@ -91,6 +91,12 @@ Name::
 	  The name of the pool.
 Total Physical::
 	  The physical usage statistics for the pool (Total / Used / Free).
+Properties::
+          Boolean valued properties that the pool may have. Each property has
+          a two-letter camel-case code. If the pool does not have the
+          property, a '~', for negation, is prepended to the property code.
+          The property codes are: Ca - indicates the pool has a cache,
+          Cr - indicates the pool is encrypted.
 
 FIELDS for stratis filesystem list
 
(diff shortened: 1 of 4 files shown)


mulkieran:
Hi all, this PR isn't complete; I'm soliciting comments on the format that is being used to show the properties. I think that we should definitely prefer this to separate columns for encrypted and has cache. I chose two letter codes, so that they would be slightly more mnemonic than single letter ones. I chose ~ for negation, because it is easier to see than a letter. I chose to put the symbol with a negation, rather than leave blank if the property was false, because it would be easier for me to read when scanning multiple pools. I expect that the codes will always be ordered alphabetically.

mulkieran:
The whole string can also be right aligned, if anybody thinks that would look better.

mulkieran:
The output looks like this now:

```
[root@localhost stratis-cli]# PYTHONPATH=./src ./bin/stratis pool list
Name                  Total Physical   Properties
name         512 EiB / 0 B / 512 EiB      ~Ca,~Cr
with_cache   512 EiB / 0 B / 512 EiB       Ca,~Cr
```

PR #490, opened 2020-03-02
Use r1 of the pool interface

mulkieran:


mulkieran:
I think it's reasonable that our blackbox test infrastructure should rely on the version that we are releasing.

mulkieran:
@bgurney-rh: If you want to run a test by hand, that would be great.

bgurney-rh:
I'm running a blackbox test on this now.

Issue #492, opened 2020-03-18
Remove "--invalid-name" flags from pylint calls in check.py


mulkieran:
We allow invalid names in some, of the subpackages, but not in all. This is because some have a reasonable number of invalid names. But we would probably be a lot better off if we were just consistent across all the subpackages.

mulkieran:
Closed by #498.

PR #494, opened 2020-03-20
Meet pylint naming requirements in whitebox tests

lleshchi:
Related #492


tests/whitebox/_misc.py
```diff
@@ -35,8 +35,8 @@
 try:
     _STRATISD = os.environ["STRATISD"]
 except KeyError:
-    message = "STRATISD environment variable must be set to absolute path of stratisd executable"
-    sys.exit(message)
+    _MESSAGE = "STRATISD environment variable must be set to absolute path of stratisd executable"
```
mulkieran: Here, please drop the unnecessary local ```_MESSAGE```` just use the literal string as an argument to ```sys.exit_```

mulkieran:
@lleshchi You accidentally included a .swo file in one of your commits.

mulkieran:
@lleshchi Your changes did indeed cause a different pylint check to be triggered. Go ahead and add a ```pylint: disable``` directive after the offending line. Most lkely this error wasn't triggered until you changed the name, because the module that checks for attributes initialized outside the ```__init__``` method has an heuristic that says ```setUp``` is like ```__init__``` and is therefore ok, but ```setup``` is just a normal method. You'll find this kind of thing happen occasionally with pylint, that by fixing one lint you trigger another.

mulkieran:
The linter is happy but the formatter doesn't like it. You may find that after you reformat, the linter doesn't like it either. The good order is fmt, then lint, for those reasons.

mulkieran:
Travis has passed.

PR #495, opened 2020-03-20
Ignore .swo as well as .swp files for vi

mulkieran:


mulkieran:
Actually, this is so trivial that nobody should be bothered with review.

PR #496, opened 2020-03-24
Meet pylint naming requirements in tests

### LINKED ISSUE #492: Remove "--invalid-name" flags from pylint calls in check.py

lleshchi:
Related #492

### DIFF
--- a/check.py
+++ b/check.py
@@ -32,7 +32,6 @@ arg_map = {
         "--reports=no",
         "--disable=I",
         "--disable=duplicate-code",
-        "--disable=invalid-name",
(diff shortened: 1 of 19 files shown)


mulkieran:
The Travis test completed succesfully, but again did not report back to GitHub. If this continues tomorrow we had better investigate it.

mulkieran:
Will also need to be done for stratis-cli/src, but that should be a different PR.

PR #501, opened 2020-04-07
Fix create_pool to make blockdevs a frozenset

### LINKED ISSUE #500: Not all invocations of _check_same_tier pass in the correct type for to_be_added
Some uses of _check_same_tier pass in lists instead of frozensets.

jbaublitz:
Closes #500

### DIFF
--- a/src/stratis_cli/_actions/_top.py
+++ b/src/stratis_cli/_actions/_top.py
@@ -157,7 +157,7 @@ class TopActions:
         managed_objects = ObjectManager.Methods.GetManagedObjects(proxy, {})
         pool_name = namespace.pool_name
         names = pools(props={"Name": pool_name}).search(managed_objects)
-        blockdevs = namespace.blockdevs
+        blockdevs = frozenset(namespace.blockdevs)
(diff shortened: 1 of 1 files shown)


jbaublitz:
This may fail CI and I know how to fix it if it does.

jbaublitz:
> This may fail CI and I know how to fix it if it does.

@mulkieran The reason I mentioned this is we pass `blockdevs` into the D-Bus call after this instead of `namespace.blockdevs`. CI seems to be okay with this but do you have any concerns about that?

mulkieran:
> > This may fail CI and I know how to fix it if it does.
> 
> @mulkieran The reason I mentioned this is we pass `blockdevs` into the D-Bus call after this instead of `namespace.blockdevs`. CI seems to be okay with this but do you have any concerns about that?

That's interesting. The ```into_dbus_python``` library takes that frozenset, relistifies it, and sticks it in a dbus-python Array. I'll go ahead and merge this; I think that into_dbus_python behavior is exactly correct.

PR #518, opened 2020-04-13
Extend the StratisCliIncoherenceError for blockdevs

mulkieran:


mulkieran:
Note that this is tested against the sim engine, but testing may not have covered all paths.

Issue #510, opened 2020-04-09
Extend blackbox tests for encryption


mulkieran:
I think this can go in a few steps. The general idea would be to work on the stratisd tests first, and then on the CLI tests.

For stratisd:
- [x] Modify testlib/dbus.py to use the new interfaces. It is already using the new pool interface, and I don't think will need any new modifications once #506 is in. So, the only modification is to the manager interface. Hopefully, it will be possible here to keep the tests in stratisd_cert.py entirely the same.
- [ ] Extend the tests in stratisd_cert.py so that they also test encryption. This should be more challenging than the previous task, because it requires keyring management to make the tests run.
The alternative is to prove that the testing infrastructure that the tests must run on will not support keyring management, and to give up entirely. We must check into this first.
- [ ] If it has been proven to be possible to do keyring management in the blackbox tests, then the CLI tests should also be extended for encryption.

Aside: if there is a global change to the test infrastructure, either for stratisd or for the CLI to make the tests have more useful message when they fail, that should be considered.

It would  be quite suitable to do each of these changes in a separate PR.

PR #525, opened 2020-04-15
stratisd encryption blackbox test

jbaublitz:
Related to #510

### DIFF
--- a/tests/blackbox/testlib/utils.py
+++ b/tests/blackbox/testlib/utils.py
@@ -15,10 +15,12 @@
 Utility functions for blackbox testing.
 """
 # isort: STDLIB
+import base64
 import os
 import random
+import shutil
 import string
-from subprocess import PIPE, Popen
+from subprocess import PIPE, Popen, run
 
 # isort: THIRDPARTY
 import psutil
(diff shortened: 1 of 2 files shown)


jbaublitz:
@mulkieran This is ready for initial review. Do we also want a pool destroy test for encrypted volumes?

tests/blackbox/testlib/utils.py
```diff
@@ -123,3 +125,67 @@ def exec_test_command(cmd):
...
+                "Command '%s' failed with exit code %s" % (" ".join(args), return_code)
+            )
+
+    def __enter__(self):
+        with open("/dev/urandom", "rb") as urandom_f:
+            key_desc = base64.b64encode(urandom_f.read(16)).decode("utf-8")
+
+        args = ["keyctl", "get_persistent", "@s", "0"]
+        exit_values = run(args, capture_output=True, text=True)
+        persistent_id = exit_values.stdout.strip()
```
mulkieran, line 170: Please swap these two lines; dosn't make sense to get ```persistent_id``` if call has failed.

tests/blackbox/testlib/utils.py
```diff
@@ -123,3 +125,67 @@ def exec_test_command(cmd):
...
+        :param args: The command line that caused the command to fail
+        :type args: list of str
+        :raises RuntimeError
+        """
+        if return_code != 0:
+            raise RuntimeError(
+                "Command '%s' failed with exit code %s" % (" ".join(args), return_code)
+            )
+
+    def __enter__(self):
```
mulkieran, line 164: Please add header docs here, especially about returned value.

tests/blackbox/testlib/utils.py
mulkieran: Actually, do you even need ```persistent_id```? man page indicates that you could use ```@s```.

jbaublitz: I'll do a test to see if this is necessary, but this comes down to how we'd recommend users administer the keychain. I'll present both options once I have more information, but the way this is currently operating, it does work so the only benefit we will be getting out of this is a slightly simpler set of commands to recommend (which may still be worth it).

jbaublitz: It appears we do need `persistent_id`.

jbaublitz:
Blocked by stratis-storage/stratisd#1961

tests/blackbox/testlib/utils.py
jbaublitz: Resolved.

tests/blackbox/testlib/utils.py
jbaublitz: Resolved.

bgurney-rh:
I tried another blackbox test run with these commits after stratis-storage/stratisd#1961 was merged, and everything passes.

jbaublitz:
@bgurney-rh Excellent! I was hoping it would match up with my local testing.

PR #533, opened 2020-04-17
Add CLI blackbox tests

jbaublitz:
Related to #510 

One note for @mulkieran: I added a test to ensure that the cache is mutually exclusive with encryption but I can remove it if that doesn't add any additional value or if the blackbox tests should not be checking this.

@bgurney-rh I was unable to test these changes from my machine so would you be able to run these after I get confirmation from @mulkieran that both added tests should be there?

### DIFF
--- a/tests/blackbox/stratis_cli_cert.py
+++ b/tests/blackbox/stratis_cli_cert.py
@@ -194,8 +194,8 @@ class StratisCertify(unittest.TestCase):  # pylint: disable=too-many-public-meth
             self.unittest_command(
                 [STRATIS_CLI, "pool", "init-cache", pool_name, StratisCertify.DISKS[2]],
                 1,
-                True,
                 False,
+                True,
             )
 
     def test_pool_list_not_empty(self):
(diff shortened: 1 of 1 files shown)


mulkieran:
@bgurney-rh Do you want to go ahead and test this? It all looks good to me.

tests/blackbox/stratis_cli_cert.py
```diff
@@ -148,6 +162,42 @@ def test_pool_create(self):
...
+        Test creating an encrypted pool with cache.
+
+        This should fail.
+        """
+        with KernelKey("test-password") as key_desc:
+            pool_name = make_test_pool(StratisCertify.DISKS[0:2], key_desc)
+            self.unittest_command(
+                [STRATIS_CLI, "pool", "init-cache", pool_name, StratisCertify.DISKS[2]],
+                1,
+                True,
```
mulkieran, line 197: For some reason stderr comes before standout in arg order and I think it's stderr that we should expect to be non-empty.

jbaublitz: Okay I assumed it would be the opposite! I'll fix that.

jbaublitz: Should be fixed now.

bgurney-rh:
@jbaublitz after trying commits 55ef10 and 78bb53, all blackbox tests pass.

PR #534, opened 2020-04-17
blackbox: change helper asserts to RuntimeErrors

bgurney-rh:
Related: stratis-storage/project#105

### DIFF
--- a/tests/blackbox/stratis_cli_cert.py
+++ b/tests/blackbox/stratis_cli_cert.py
@@ -34,6 +34,20 @@ from testlib.utils import (
 )
 
 
+def _raise_error_exception(return_code, msg):
+    """
+    Check result of a D-Bus call in a context where it is in error
+    if the call fails.
+    :param int return_code: the return code from the D-Bus call
+    :param str msg: the message returned on the D-Bus
+    """
+    if return_code != 0:
+        raise RuntimeError(
+            "Expected return code of 0; actual return code: %s, error_msg: %s"
+            % (return_code, msg)
+        )
+
+
 def make_test_pool(pool_disks, key_desc=None):
     """
     Create a test pool that will later get destroyed
@@ -51,11 +65,8 @@ def make_test_pool(pool_disks, key_desc=None):
     args += [pool_name] + pool_disks
 
     (return_code, _, stderr) = exec_test_command(args)
-    if return_code != 0:
-        raise RuntimeError(
-            "Expected return code of 0; actual return_code: %s, %s"
-            % (return_code, stderr)
-        )
+
+    _raise_error_exception(return_code, stderr)
     return pool_name
 
 
@@ -69,11 +80,8 @@ def make_test_filesystem(pool_name):
     (return_code, _, stderr) = exec_test_command(
         [STRATIS_CLI, "filesystem", "create", pool_name, filesystem_name]
     )
-    if return_code != 0:
-        raise RuntimeError(
-            "Expected return code of 0; actual return_code: %s, %s"
-            % (return_code, stderr)
-        )
+
+    _raise_error_exception(return_code, stderr)
     return filesystem_name
 
 
--- a/tests/blackbox/stratis_cli_cert.py
+++ b/tests/blackbox/stratis_cli_cert.py
@@ -51,7 +51,11 @@ def make_test_pool(pool_disks, key_desc=None):
     args += [pool_name] + pool_disks
 
     (return_code, _, stderr) = exec_test_command(args)
-    assert return_code == 0, "return_code: %s, stderr: %s" % (return_code, stderr)
+    if return_code != 0:
+        raise RuntimeError(
+            "Expected return code of 0; actual return_code: %s, %s"
+            % (return_code, stderr)
+        )
     return pool_name
 
 
@@ -65,7 +69,11 @@ def make_test_filesystem(pool_name):
     (return_code, _, stderr) = exec_test_command(
         [STRATIS_CLI, "filesystem", "create", pool_name, filesystem_name]
     )
-    assert return_code == 0, "return_code: %s, stderr: %s" % (return_code, stderr)
+    if return_code != 0:
+        raise RuntimeError(
+            "Expected return code of 0; actual return_code: %s, %s"
+            % (return_code, stderr)
+        )
     return filesystem_name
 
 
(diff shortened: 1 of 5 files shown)


tests/blackbox/stratis_cli_cert.py
```diff
@@ -37,7 +37,11 @@ def make_test_pool(pool_disks):
     (return_code, _, stderr) = exec_test_command(
         [STRATIS_CLI, "pool", "create", pool_name] + pool_disks
     )
-    assert return_code == 0, "return_code: %s, stderr: %s" % (return_code, stderr)
+    if return_code != 0:
+        raise RuntimeError(
+            "Expected return code of 0; actual return_code: %s, %s"
+            % (return_code, stderr)
+        )
```
jbaublitz, line 44: Is it worth factoring these blocks into a method? It looks like it shows up a few places and it's just barely more than trivial so it might be worth keeping that consistent in a method.

tests/blackbox/stratisd_cert.py
```diff
@@ -37,8 +37,13 @@ def make_test_pool(pool_name, pool_disks):
     (obj_path_exists, (obj_path, _)), return_code, msg = StratisDbus.pool_create(
         pool_name, pool_disks, None
     )
-    assert return_code == 0, "return_code: %s, error_msg: %s" % (return_code, msg)
-    assert obj_path_exists, "obj_path_exists: %s" % obj_path_exists
+    if return_code != 0:
```
mulkieran, line 40: I think abstracting is a good idea, but we should probably abstract separately for stratisd_cert.py and stratis_cli_cert.py.

A method something like:

```
def _raise_error_exception(return_code, msg, return_value_exists):
    """
    Check result of a D-Bus call in a context where it is in error if the call fails.
    
    :param int return_code: the return code from the D-Bus call
    :param str msg: the message returned on the D-Bus
    :param bool return_value_exists: whether a value representing a valid result was returned
    """
    if return_code != 0:
       raise RuntimeError(
           "Expected return code of 0; actual return_code: %s, error_msg: %s" % (return_code, msg)
      )

    if not return_value_exists:
       raise RuntimeError("result value was default or place holder value and does not represent a valid result")
```

then invoke here as:

```
_raise_error_exception(return_code, msg, obj_path_exists)
```

would convey all the useful information.

tests/blackbox/stratis_cli_cert.py
mulkieran: We might as well abstract for stratis_cli_cert.py as well, while we're here, even though these checks are smaller...

mulkieran:
@bgurney-rh Best to resolve conflicts in stratis_cli_cert.py now. They arise because of the added parameter for encryption.

bgurney-rh:
Rebased and resolved conflict.

tests/blackbox/stratisd_cert.py
bgurney-rh: Resolved.

tests/blackbox/stratis_cli_cert.py
```diff
@@ -34,6 +34,20 @@
 )
 
 
+def _raise_error_exception(return_code, msg):
+    """
+    Check result of a D-Bus call in a context where it is in error
```
mulkieran, line 39: Please fix up the header comment so that it is correct for this context: using the CLI rather than directly accessing the D-Bus.

bgurney-rh: Resolved.

tests/blackbox/stratis_cli_cert.py
bgurney-rh: Resolved.

bgurney-rh:
I converted the testlib assert calls into RuntimeErrors.  I wasn't sure of what messages should be added (if any), so I made a guess.

tests/blackbox/testlib/dbus.py
```diff
@@ -406,4 +406,5 @@ def clean_up():
     :return: None
     """
     StratisDbus.destroy_all()
-    assert StratisDbus.pool_list() == []
+    if StratisDbus.pool_list() != []:
+        raise RuntimeError("testlib dbus: clean_up failed")
```
mulkieran, line 410: You might as well report the names of the pool's that weren't destroyed here.

tests/blackbox/testlib/stratis.py
```diff
@@ -83,4 +83,5 @@ def clean_up():
     :return: None
     """
     _StratisCli.destroy_all()
-    assert _StratisCli.pool_list() == []
+    if _StratisCli.pool_list() != []:
+        raise RuntimeError("testlib cli: clean_up failed")
```
mulkieran, line 87: You can report the names here as well.

tests/blackbox/testlib/utils.py
```diff
@@ -107,6 +107,8 @@ def exec_command(cmd):
         print("STDERR= %s" % stderr_text)
 
     assert expected_exit_code == exit_code
+    if expected_exit_code != exit_code:
```
mulkieran, line 110: For here, it makes sense to stow all the information that is in those print statements above this line in the RuntimeError, and to get rid of the print statements entirely.

bgurney-rh: I forgot to remove the "assert" line above; fixed that, and added the print statements.

bgurney-rh: Resolved.

tests/blackbox/testlib/stratis.py
bgurney-rh: Resolved.

tests/blackbox/testlib/dbus.py
bgurney-rh: Resolved.

tests/blackbox/testlib/utils.py
```diff
@@ -95,18 +95,17 @@ def exec_command(cmd):
...
 
     expected_exit_code = 0
 
     if expected_exit_code != exit_code:
-        print("cmd = %s [%d != %d]" % (str(cmd), expected_exit_code, exit_code))
-        print("STDOUT= %s" % stdout_text)
-        print("STDERR= %s" % stderr_text)
-
-    assert expected_exit_code == exit_code
+        raise RuntimeError(
```
mulkieran, line 105: You can get rid of ```expected_exit_code``` entirely now,  since it's only used in one place.

bgurney-rh: Resolved.

PR #535, opened 2020-04-17
Add CLI blackbox tests

### LINKED ISSUE #510: Extend blackbox tests for encryption

jbaublitz:
Closes #510

### DIFF
--- a/tests/blackbox/stratis_cli_cert.py
+++ b/tests/blackbox/stratis_cli_cert.py
@@ -24,19 +24,33 @@ import unittest
 
 # isort: THIRDPARTY
 from testlib.stratis import STRATIS_CLI, clean_up
-from testlib.utils import exec_command, exec_test_command, fs_n, p_n, process_exists
+from testlib.utils import (
+    KernelKey,
+    exec_command,
+    exec_test_command,
+    fs_n,
+    p_n,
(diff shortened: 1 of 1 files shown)


mulkieran:
@jbaublitz Please feel free to merge this once you've rebased.

Issue #509, opened 2020-04-09
Remove two "no cover" pragmas


mulkieran:
There are two that can certainly be removed.

```TopActions.add_cache_devices``` has a pragma on ```return_code != StratisdErrors.OK```, but that is now unnecessary. Now that we're using the new pool interface it is possible to get an engine error by calling this command without having previously called the cache initialize command.

```TopActions.init_cache``` has a pragma on ```return_code != StratisdErrors.OK```, but that is unnecessary. You can force an error by creating an encrypted pool and then initializing the cache. Since these tests run against the sim engine, you cause an encrypted pool to be created just by passing the "--encrypted" flag with some dummy key description when you create the pool, you don't have to worry about manipulating the kernel keyring to manage the test.

mulkieran:
Blocked by https://github.com/stratis-storage/project/issues/166.

mulkieran:
Closed by attached issue.

PR #539, opened 2020-04-23
Update cache tests for new interface

lleshchi:
Related #509
Expect some test failure until stratis-storage/stratisd#1969 is merged.

### DIFF
--- a/tests/whitebox/integration/pool/test_add.py
+++ b/tests/whitebox/integration/pool/test_add.py
@@ -103,7 +103,9 @@ class AddDataTestCase1(SimTestCase):
         an exception.
         """
         devices = _DEVICE_STRATEGY()
-        command_line = ["--propagate", "pool", "add-cache"] + [self._POOLNAME] + devices
+        command_line = (
+            ["--propagate", "pool", "init-cache"] + [self._POOLNAME] + devices
+        )
         RUNNER(command_line)
         self.check_error(
             StratisCliInUseOtherTierError,
@@ -117,7 +119,9 @@ class AddDataTestCase1(SimTestCase):
         an exception.
         """
         devices = _DEVICE_STRATEGY_2()
-        command_line = ["--propagate", "pool", "add-cache"] + [self._POOLNAME] + devices
+        command_line = (
+            ["--propagate", "pool", "init-cache"] + [self._POOLNAME] + devices
+        )
         RUNNER(command_line)
         self.check_error(
             StratisCliInUseOtherTierError,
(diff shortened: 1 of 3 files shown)


tests/whitebox/integration/pool/test_add.py
```diff
@@ -51,19 +52,23 @@ def test_add(self):
...
-    Test adding devices to a non-existant pool.
+    Test adding cache devices before initializing cache.
     """
 
     _MENU = ["--propagate", "pool", "add-cache"]
     _POOLNAME = "deadpool"
 
+    def setUp(self):
+        super().setUp()
+        command_line = ["pool", "create", self._POOLNAME] + _DEVICE_STRATEGY()
```
jbaublitz, line 63: We need the `--propagate` flag here. That's why CI is failing.

jbaublitz: Disregard this. I need to do a little more digging.

jbaublitz:
If your fix for the sim engine hasn't been merged yet, that would explain why this is failing.

lleshchi:
Other possible tests:
-  Add devices to the cache of pool1, then add same devices to pool2
- Testing `add-cache` and `add-data` succeed when adding multiple devices
- Testing `add-cache` and `add-data` succeed when adding zero devices
- Initialize cache, add device to cache tier with `add-cache`, then attempt to add the device to the data tier with `add-data `(This might be redundant. There is a test to `init-cache` with a device, and then `add-data` with the same device.)
- Initialize pool, add device with `add-data`, then attempt to add the device to cache (This might also be redundant, since there is a test which initializes a pool with a device, then attempts to add that device to the cache)

lleshchi:
Additionally, there are currently only 4 tests for` init-cache`. Many of the `add-cache` tests can be applied to `init-cache`, such as testing interactions between `init-cache` and `add-data`.

mulkieran:
The tests look good so far. If you think there are more tests that are desirable to add go right ahead. Please add them individually and make sure you run each new test against your private copy of stratisd, to verify that all is as expected.

mulkieran:
All the additional changes look fine to me. If you feel satisfied with the tests, and you've verified that the tests and coverage are passing against your private version of stratisd, then we should consider this ready.

Please add a link to the corresponding stratisd PR.

You can decide whether you want to squash some commits or not. I'm happy with them as they are. You might want to fix up a few commit messages, though, there's a typo: ```deviced```.

mulkieran:
@lleshchi Please rebase.

PR #544, opened 2020-04-25
Blackbox test for Report interface

### LINKED ISSUE #540: Add a test of report interface to the blackbox tests
This should be _very_ simple indeed, but it might as well verify that the keyword works, and that the result is valid JSON. No need to be so clever as to make the report other than an empty list. It should only be on the stratisd side right now, as we do not currently make use of it in the CLI.

jbaublitz:
Closes #540


mulkieran:
We'll keep the CLI issue quite separate, so I"ll say this closes #540.

PR #545, opened 2020-04-27
Use Python 3.6.8 equivalents for some Python 3.7 conveniences

mulkieran:


mulkieran:
The reason that our Travis pylint task didn't catch these seems to be a pylint issue rather than a task issue. I verified that our lint task does in fact lint these files and that the version of Python that is being run seems to be correct.

bgurney-rh:
Successfully tested via the blackbox tests on a machine with Python 3.6.8 installed.

mulkieran:
@bgurney-rh thx

PR #547, opened 2020-04-28
Future-proof Python code

jbaublitz:
Related to stratis-storage/stratisd#1989.

### DIFF
--- a/src/stratis_cli/__init__.py
+++ b/src/stratis_cli/__init__.py
@@ -14,6 +14,6 @@
 """
 Top level of CLI.
 """
-from ._main import run
 from ._errors import StratisCliEnvironmentError
 from ._exit import StratisCliErrorCodes, exit_
+from ._main import run
--- a/src/stratis_cli/_actions/__init__.py
+++ b/src/stratis_cli/_actions/__init__.py
@@ -15,9 +15,7 @@
 Package mediating dbus actions.
 """
 
-from ._constants import BLOCKDEV_INTERFACE
-from ._constants import FILESYSTEM_INTERFACE
-from ._constants import POOL_INTERFACE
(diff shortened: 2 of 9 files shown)


jbaublitz:
@mulkieran I'm a little confused at the return of the non-top level import linting errors. Do you know why this is?

mulkieran:
Likely when we were running under tox we were running an older version of Python but actually a newer version of pylint. When we switched to what we are running now, we caught some actual errors, presumably because the newer version of Python and the older version of pylint had some better interaction, so it looked like an upgrade all around, but it wasn't actually.

jbaublitz:
@mulkieran I tracked down the Github issue where they added the import check that is breaking CI. This means that this PR will only be relevant when we decide to update pylint in CI.

jbaublitz:
@mulkieran @bgurney-rh While I can update this easily in Travis, how would you like me to proceed for Jenkins? I don't think we can't put this through until we update the Jenkins worker nodes, but I'm not sure if there's another way to handle that.

mulkieran:
@jbaublitz: We can't put this through until @lleshchi  and I are ready to move up a Fedora version. I don't think it'll matter to the Jenkins nodes, but I expect that @bgurney-rh will be in process of updating those soon or is already.

mulkieran:
An alternative is to do a dynamic check for the pylint version that is being run against, and then we can work with both pylint versions. I'm going to give that a go, especially as recent changes have made this PR pretty stale.

mulkieran:
> An alternative is to do a dynamic check for the pylint version that is being run against, and then we can work with both pylint versions. I'm going to give that a go, especially as recent changes have made this PR pretty stale.

Alas, it's not just pylint, it's also black that causes the problem, since the new version formats one line differently. So, we'll just have to halt for a bit longer.

jbaublitz:
Out of date. Closing and waiting until we migrate to Fedora 31.

PR #548, opened 2020-04-29
Add key operations to CLI

jbaublitz:
Putting up for initial review.

Related to stratis-storage/project#171

One note: I can see us potentially getting questions about why we provide information about whether an existing key that is being reset has changed or not. This type of question comes up a lot (whether you are leaking information or not). Here, I'd argue we are not leaking information as the user needs root permissions to access that information from stratisd and otherwise they will get no information whatsoever. As long as we keep operations that add keys root-only, our threat model shouldn't change at all. We should however consider the implications of this if we ever decide that only certain Stratis users should have access to certain keys (an issue that comes up when discussing access control schemes). If we do implement that at some point, this will become relevant.

TODO: Add tests for full coverage.


bgurney-rh:
The only CI failures appear to be coverage:
```
Name                                      Stmts   Miss Branch BrPart  Cover   Missing
-------------------------------------------------------------------------------------
src/stratis_cli/_actions/_top.py            182     45     46      0    72%   152-172, 188-207, 623-629, 665-680, 702-710, 735-739
src/stratis_cli/_errors.py                   86      3     14      0    97%   58-59, 62
```

jbaublitz:
@mulkieran Holding off on applying these per stand up.

mulkieran:
blocked by #551

mulkieran:
Please rebase. You'll need to fix up ```InitCacheFail4TestCase``` to model key managment, I think, as well.

src/stratis_cli/_actions/_top.py
```diff
@@ -538,3 +602,139 @@ def add_cache_devices(namespace):  # pylint: disable=too-many-locals
...
+            raise StratisCliNameConflictError("key", namespace.keydesc)
+
+        ((changed, existing_modified), return_code, message) = add_update_key(
+            proxy,
+            namespace.keydesc,
+            namespace.capture_key,
+            keyfile_path=namespace.keyfile_path,
+        )
+
+        if return_code != StratisdErrors.OK:  # pragma: no cover
```
mulkieran, line 629: You should be able to remove the "no cover" pragma here, since the engine should reliably return an error if the key is too long.

src/stratis_cli/_actions/_top.py
```diff
@@ -538,3 +602,139 @@ def add_cache_devices(namespace):  # pylint: disable=too-many-locals
...
+            raise StratisCliResourceNotFoundError("reset", namespace.keydesc)
+
+        ((changed, existing_modified), return_code, message) = add_update_key(
+            proxy,
+            namespace.keydesc,
+            namespace.capture_key,
+            keyfile_path=namespace.keyfile_path,
+        )
+
+        if return_code != StratisdErrors.OK:  # pragma: no cover
```
mulkieran, line 674: You can remove the "no cover" pragma here because it is generally testable via the too long key.

tests/whitebox/integration/pool/test_init_cache.py
```diff
@@ -117,14 +118,16 @@ class InitCacheFail4TestCase(SimTestCase):
...
-        command_line = [
-            "pool",
-            "create",
-            "--key-desc",
-            "test-password",
-            self._POOLNAME,
-        ] + _DEVICE_STRATEGY()
-        RUNNER(command_line)
+
+        with RandomKeyTmpFile() as keyfile_path:
```
mulkieran, line 122: This is the wrong context manager, ```create pool``` only accepts a key description.

mulkieran:
Blocked by https://github.com/stratis-storage/stratisd/pull/1983.

docs/stratis.txt
```diff
@@ -66,6 +68,15 @@ filesystem rename <pool_name> <fs_name> <new_name>::
 blockdev list [pool_name]::
 	 List all blockdevs that make up the specified pool, or all pools, if
 	 no pool name is given.
+key set <(--keyfile-path | --capture-key)> <key_desc>::
```
mulkieran, line 71: Wherever there's the flag ```--keyfile-path``` it should also include the ```<keyfile_path>``` argument that it must take.

docs/stratis.txt
```diff
@@ -78,9 +89,8 @@ OPTIONS
         is also the default.
 --key-desc::
         The key description of the key that should be used to encrypt the
-        created pool. The key description must correspond to a kernel key
-        that is managed by the user and has at least one link to a searchable
-	keyring (thread, process, or session keyrings).
+        created pool. The key description must correspond to a key
```
mulkieran, line 92: Last sentence revision:

```The key may be assigned to the key description in the kernel keyring using the *key* command.```

Because it could be done via the old school keyctl approach just as well, or with other mechanisms that we might be contemplating.

Comment (jbaublitz) [reply docs/stratis.txt:92]: This is true that they could theoretically do this, but it's going to be inconvenient and even difficult for users to do it through keyctl as we add a prefix to the key that is not public. Beyond that, we can't do key length checking if the user uses keyctl. Over all, I think it's a bad idea to allow the old approach as we lose some of the guardrails we've put in place.

File: docs/stratis.txt (hunk 4 above)
Comment (jbaublitz) [reply docs/stratis.txt:71]: Oops! You're right.

File: docs/stratis.txt (hunk 5 above)
Comment (mulkieran) [reply docs/stratis.txt:92]: > This is true that they could theoretically do this, but it's going to be inconvenient and even difficult for users to do it through keyctl as we add a prefix to the key that is not public. Beyond that, we can't do key length checking if the user uses keyctl. Over all, I think it's a bad idea to allow the old approach as we lose some of the guardrails we've put in place.

@jbaublitz: Is the prefix added to the key or to the key description?

Comment (jbaublitz) [reply docs/stratis.txt:92]: It's added to the key description so the system key description displayed by keyctl will differ from what's displayed and used by Stratis at the application level. Otherwise we would not be able to tell which keys are set by Stratis and which ones are managed by other users. If we did not set this, we would theoretically allow a user to unlink a key in the keyring that is being used for something other than Stratis which I think we want to avoid.

**Comment by mulkieran** [comment]
blocked by https://github.com/stratis-storage/stratisd/pull/1983.

**Comment by mulkieran** [comment]
Looks fine, leaving in pending.

**Comment by jbaublitz** [comment]
@mulkieran Addressed the migration away from `keyctl` in the blackbox tests. We will need to test the blackbox test changes against the stratisd side PR that exposes the key API.

**Comment by mulkieran** [comment]
> @mulkieran Addressed the migration away from `keyctl` in the blackbox tests. We will need to test the blackbox test changes against the stratisd side PR that exposes the key API.

@jbaublitz: We use KernelKey in udev tests in stratisd also. So it makes sense to update your stratisd key management PR as well.

**Comment by jbaublitz** [comment]
> > @mulkieran Addressed the migration away from `keyctl` in the blackbox tests. We will need to test the blackbox test changes against the stratisd side PR that exposes the key API.
> 
> @jbaublitz: We use KernelKey in udev tests in stratisd also. So it makes sense to update your stratisd key management PR as well.

I already started working on that but it's going to take a bit more time than the CLI side because we should really be using D-Bus directly in the udev tests.

**Comment by mulkieran** [comment]
> > > @mulkieran Addressed the migration away from `keyctl` in the blackbox tests. We will need to test the blackbox test changes against the stratisd side PR that exposes the key API.
> > 
> > 
> > @jbaublitz: We use KernelKey in udev tests in stratisd also. So it makes sense to update your stratisd key management PR as well.
> 
> I already started working on that but it's going to take a bit more time than the CLI side because we should really be using D-Bus directly in the udev tests.

I'm afraid you may have to split KernelKey into two implmentations for blackbox tests, one for the CLI and one for the daemon.

**Comment by jbaublitz** [comment]
> I'm afraid you may have to split KernelKey into two implmentations for blackbox tests, one for the CLI and one for the daemon.

We aren't currently doing encryption testing in the blackbox tests so I don't think that belongs in this PR. Once we add it in another PR, we will probably have to do that.

**Comment by jbaublitz** [comment]
@mulkieran Is it worth keeping them consolidated and using the D-Bus API for both as we can always add some blackbox tests for setting the keys specifically through the CLI using the infrastructure that we've already designed for CLI commands?

**Comment by mulkieran** [comment]
> @mulkieran Is it worth keeping them consolidated and using the D-Bus API for both as we can always add some blackbox tests for setting the keys specifically through the CLI using the infrastructure that we've already designed for CLI commands?

@jbaublitz I would be fine with a single KernelKey implementation that worked directly with the D-Bus.

**Comment by mulkieran** [comment]
@jbaublitz Please rebase to resolve conflicts.

File: src/stratis_cli/_actions/_top.py [hunk 6]
```diff
@@ -558,3 +622,142 @@ def get_report(namespace):
...
+            raise StratisCliNoChangeError("remove", namespace.keydesc)
+
+        (changed, return_code, message) = Manager.Methods.DeleteKey(
+            proxy, {"key_desc": namespace.keydesc}
+        )
+
+        if return_code != StratisdErrors.OK:  # pragma: no cover
+            raise StratisCliEngineError(return_code, message)
+
+        # A new key was added even though there was a key reported to already exist.
```
mulkieran, line 739: You can drop the comment, it doesn't match the IncoherenceError message, which I think is correct.

src/stratis_cli/_actions/_top.py
```diff
@@ -136,6 +142,64 @@ def _check_same_tier(pool_name, managed_objects, to_be_added, this_tier):
...
+        file_desc = sys.stdout.fileno()
+        terminal_attributes = tcgetattr(file_desc)
+        setcbreak(file_desc)
+        fd_is_terminal = True
+        print("Enter desired key data followed by the return key:")
+    else:
+        file_desc = os.open(keyfile_path[0], os.O_RDONLY)
+        fd_is_terminal = False
+
+    add_ret = Manager.Methods.AddKey(
```
mulkieran, line 190: This should be ```SetKey``` now.

src/stratis_cli/_actions/_top.py
```diff
@@ -136,6 +142,64 @@ def _check_same_tier(pool_name, managed_objects, to_be_added, this_tier):
...
+    """
+    Issue a command to set or reset a key in the kernel keyring with the option
+    to set it interactively or from a keyfile.
+
+    :param proxy: proxy to the top object
+    :param str key_desc: key description for the key to be set or reset
+    :param bool capture_key: whether the key setting should be interactive
+    :param keyfile_path: optional path to the keyfile containing the key
+    :type keyfile_path: list of str or NoneType (if list, exactly one element)
+    :return: the result of the AddKey D-Bus call
```
mulkieran, line 173: ```AddKey``` -> ```SetKey``` here also.

src/stratis_cli/_actions/_top.py
```diff
@@ -558,3 +622,142 @@ def get_report(namespace):
...
+        """
+        from ._data import Manager
+
+        proxy = get_object(TOP_OBJECT)
+
+        key_list = _fetch_keylist_property(proxy)
+        if namespace.keydesc not in key_list:
+            raise StratisCliNoChangeError("remove", namespace.keydesc)
+
+        (changed, return_code, message) = Manager.Methods.DeleteKey(
```
mulkieran, line 732: ```DeleteKey``` -> ```UnsetKey```

PR #549, opened 2020-04-30
Jbaublitz issue project 171

mulkieran:


jbaublitz:
@mulkieran With the last commit changed to `nargs=1`, have you verified that this does not return a list? That was the reason I originally set it to `nargs="?"`. The documentation seems to indicate that `nargs=1` will return a list of a single element.

mulkieran:
> @mulkieran With the last commit changed to `nargs=1`, have you verified that this does not return a list? That was the reason I originally set it to `nargs="?"`. The documentation seems to indicate that `nargs=1` will return a list of a single element.

It _will_ return a list. Guess I'll have to do a more holistic review. But I think it will be best to accept some complication on the inside rather than confusion on the outside.

mulkieran:
Merged into #548.

PR #550, opened 2020-05-01
Rework getting properties from FetchProperties result

mulkieran:
This changes the method used to get a property from the result of a FetchProperties call for various listing methods, fixing a bunch of little mistakes that had crept in, and making the list code a bit more succinct.

It also exposes a revised ```fetch_properties``` method, which is expected to be used in an upcoming PR.

### DIFF
--- a/src/stratis_cli/_actions/_formatting.py
+++ b/src/stratis_cli/_actions/_formatting.py
@@ -18,7 +18,7 @@ Formatting for tables.
 # isort: STDLIB
 import sys
 
-from .._errors import StratisCliPropertyNotFoundError
+from ._utils import fetch_property
 
 # If the wcwidth package is not available the wcswidth function will not
 # be available. In that case, use the standard function len where wcswidth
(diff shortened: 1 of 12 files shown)


mulkieran:
Related: #548 .

mulkieran:
Whoops. One more adjustment.

mulkieran:
I forgot to push the test changes in the last commit. Now in with the rest.

src/stratis_cli/_actions/_formatting.py
```diff
@@ -46,35 +46,26 @@
...
-        if not success:
-            return None  # pragma: no cover
-        return to_repr(variant)
-    except KeyError:  # pragma: no cover
-        raise StratisCliPropertyNotFoundError(object_type, name)
+        return to_repr(fetch_property(props, name))
+    # An exception should only be raised if the property can not be obtained.
+    # This requires either running against an interface that does not support
+    # the property or the engine encountering an error getting the property.
+    except:  # pylint: disable=bare-except # pragma: no cover
```
jbaublitz, line 67: Should we be doing a bare except here? If so, what is the rationale for this? This would swallow a `SIGINT` sent from the terminal if sent during this block which is a bit of a departure from what we do elsewhere.

mulkieran: Yeah. That's going too far. Let me tidy it up.

mulkieran:
coverage, one more time...

jbaublitz:
I think coverage is due to a typo here: https://github.com/stratis-storage/stratis-cli/pull/550/files#diff-631b946c80beffd10afcb7922ae8c46cR72

Issue #514, opened 2020-04-09
Should we revive the check for compatible stratisd version 


mulkieran:
We had an issue for this, #204, and we implemented a solution. But the project wasn't organized enough to keep the required version information properly up-to-date, so the merged code had to be reverted.

It might be worth revisiting the idea again, now that we can potentially get odd or confusing errors by running against a too old stratisd version.

mulkieran:
We can by:

* Adding a new exception, say ```StratisCliBadStratisVersionError``` which is a direct subtype of ```StratisCliRuntimeException```.
* Restoring the small bit of the introspection data of the last Manger interface so that the Version property can be accessed
* Checking in the ```run()``` method after parsing but before any action is executed and raising the error of the version of straitsd doesn't check out. This will require one D-Bus call, to get the Version value as reported by the old Manager interface.
* Treating the raised exception normally and handling it in error_reporting.py as appropriate.

mulkieran:
We decided to go ahead with this in standup today.

PR #552, opened 2020-05-07
Check for stratisd version compatibility

lleshchi:
Related #514 
Supersedes https://github.com/stratis-storage/stratisd/pull/2025

Expect tests to fail until the stratisd version update is merged


src/stratis_cli/_actions/_stratisd_version.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    Checks that the version of stratisd that is running is compatible with
+    this version of the CLI
+
+    :raises StratisCliStratisdVersionError
+    """
+    from ._connection import get_object
+    from ._constants import TOP_OBJECT
+    from ._data import MAXIMUM_STRATISD_VERSION, MINIMUM_STRATISD_VERSION, Manager
+    from .._errors import StratisCliStratisdVersionError
```
jbaublitz, line 16: Only the `._data` import needs to be loaded inside the method. The rest can be top level imports.

lleshchi:
Should I add no cover pragmas to the lines which are failing coverage?

mulkieran:
> Should I add no cover pragmas to the lines which are failing coverage?

Yes. Go ahead and mark them now, so that the CI passes, and then we''ll work on ways to drop them.

mulkieran:
It turns out that the existing monkey patching example is very helpful...I actually patched the Version Get method, because it was about the easiest thing to patch for the thing I was trying to test. So you can probably work to adapt that test fairly easily.

tests/whitebox/monkey_patching/test_stratisd_version.py
```diff
@@ -0,0 +1,52 @@
+# Copyright 2016 Red Hat, Inc.
```
mulkieran, line 1: You'll want to change the year.

src/stratis_cli/_actions/_data.py
```diff
@@ -36,6 +36,11 @@
 )
 from ._utils import get_timeout
 
+MAXIMUM_STRATISD_VERSION = (2, 0, 1)
```
mulkieran, line 39: Can these constants and their accompanying assertion go in ```_constants.py```?

src/stratis_cli/_parser/_parser.py
```diff
@@ -55,7 +61,17 @@ def add_subcommand(subparser, cmd):
...
+        if func is None:
+            return PRINT_HELP(parser)
+
+        def wrapped_func(*args):
+            check_stratisd_version()
+            func(*args)
+
+        return wrapped_func
+
+    parser.set_defaults(func=wrap_func(info.get("func", None)))
```
mulkieran, line 74: The default default for ```get``` is ```None```, so you shouldn't pass it explicitly.

mulkieran:
I remembered that you'll want to put up a PR against the ```cli.sh``` file in the ```ci``` repo, since you're adding a new distinct test.

mulkieran:
Ok! Go ahead with the use of the semantic_version package.

src/stratis_cli/_actions/_stratisd_version.py
```diff
@@ -0,0 +1,43 @@
...
+    Checks that the version of stratisd that is running is compatible with
+    this version of the CLI
+
+    :raises StratisCliStratisdVersionError
+    """
+    from ._constants import MAXIMUM_STRATISD_VERSION, MINIMUM_STRATISD_VERSION
+    from ._data import Manager
+
+    version_spec = SimpleSpec(
+        ">=%s-,<%s-" % (MINIMUM_STRATISD_VERSION, MAXIMUM_STRATISD_VERSION)
```
lleshchi, line 36: I believe this spec string could be moved into ._data.py or ._constants.py

mulkieran:
@lleshchi: You should include the package in the dependencies in setup.py. Also, please ask @bgurney-rh to update the environment on his test machines.

src/stratis_cli/_actions/_constants.py
```diff
@@ -24,3 +26,7 @@
 FILESYSTEM_INTERFACE = "org.storage.stratis2.filesystem"
 POOL_INTERFACE = "org.storage.stratis2.pool.r1"
 BLOCKDEV_INTERFACE = "org.storage.stratis2.blockdev"
+
+MAXIMUM_STRATISD_VERSION = "2.1.0"
+MINIMUM_STRATISD_VERSION = "2.0.0"
+assert Version(MINIMUM_STRATISD_VERSION) < Version(MAXIMUM_STRATISD_VERSION)
```
lleshchi, line 32: Which min and max versions should be accepted currently? 
Pre-release versions cannot be specified here due to the format of the spec string https://github.com/stratis-storage/stratis-cli/pull/552/files#diff-c9e6aa089520ef1da260e2c79d1d9e8fR36. This can be revised.

mulkieran: > Which min and max versions should be accepted currently?
> Pre-release versions cannot be specified here due to the format of the spec string https://github.com/stratis-storage/stratis-cli/pull/552/files#diff-c9e6aa089520ef1da260e2c79d1d9e8fR36. This can be revised.

Let's talk about this in standup.

.travis.yml
```diff
@@ -29,6 +29,7 @@ jobs:
           python-dateutil==2.8.0
           wcwidth==0.1.7
           psutil==5.4.3
+          semantic_version==2.8.5
```
mulkieran, line 32: Go with version = 2.6.0, that's the default on my system. We'll bump them all when we move up to Fedora 31.

bgurney-rh: Agreed; 2.6.0 will be what's installed on the test systems now.

(I'm re-running the test after installing it.)

bgurney-rh: I'm getting this failure:
```
ImportError: cannot import name 'SimpleSpec' from 'semantic_version' (/usr/lib/python3.7/site-packages/semantic_version/__init__.py)
```

mulkieran: Ok. Good catch. We will have to be 2.6.0 compatible, regardless of our recommended development version. It looks like there's some prior version alternatives to SimpleSpec.

lleshchi: I believe SimpleSpec was introduced in version 2.7.0

lleshchi: Replacing SimpleSpec with Spec should fix this. Testing now

src/stratis_cli/_actions/_stratisd_version.py
mulkieran: > I believe this spec string could be moved into _data.py or _constants.py
> The "-" allows for the inclusion of pre-release versions. This may not be necessary at the moment.

We decided in stand up to ignore the possibility of pre-release versions for not. It is probably best to leave the spec string right where it is, since it is so specific to this particular library.

src/stratis_cli/_actions/_stratisd_version.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def check_stratisd_version():
+    """
+    Checks that the version of stratisd that is running is compatible with
+    this version of the CLI.
+
+    :raises StratisCliStratisdVersionError
+    """
+    from ._constants import MAXIMUM_STRATISD_VERSION, MINIMUM_STRATISD_VERSION
```
mulkieran, line 32: Move the ```_constants``` import up to the top, with the other ```_constants``` import.

PR #556, opened 2020-05-13
Do test discovery via unitttest discover

mulkieran:
Related https://github.com/stratis-storage/project/issues/105

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PYTEST_OPTS = --verbose
+UNITTEST_OPTS = --verbose
 
 .PHONY: lint
 lint:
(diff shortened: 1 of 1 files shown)


mulkieran:
Note that if you check over the Jenkins test output you'll see that with the pure unittest approach it is quite verbose. I'm not sure how I feel about that. Sometimes it does us good to see the output of our own programs. On the other hand, it might be rather a lot to scroll through. To get rid of it would require somewhat clever programming, probably around the ```RUNNER``` method, which is now just ```run()```, but which could use a context manager to redirect stdout and stderr and invoke the ```run()``` result within the context manager.

mulkieran:
Sending standard output to ```/dev/null``` looks like it would be helpful, so I'm trying that.

PR #563, opened 2020-05-18
Bump required version of justbytes to 0.14

mulkieran:


mulkieran:
This might not break any of our CI, and it might not break any existing setups, either. But the requirement in the setup file is satisfiable only via pip OR dnf on rawhide.

PR #564, opened 2020-05-19
Fix D-Bus calls in KernelKey context manager

mulkieran:


mulkieran:
@bgurney-rh Please test this when you get the chance.

mulkieran:
The force push got rid of a stray file, ```junk.tx``` which had been included accidentally.

mulkieran:
This is tested now and I expect it to pass.

PR #566, opened 2020-05-19
Match GetReport with stratisd changes

jbaublitz:
Closes stratis-storage/project#181

### DIFF
--- a/tests/blackbox/testlib/dbus.py
+++ b/tests/blackbox/testlib/dbus.py
@@ -405,7 +405,7 @@ class StratisDbus:
         Get the report with the given name.
         :param str report_name: The name of the report
         :return: The JSON report as a string with a status code and string
-        :rtype: The D-Bus types (bs), q, and s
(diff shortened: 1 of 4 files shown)


tests/blackbox/testlib/dbus.py
```diff
@@ -405,7 +405,7 @@ def get_report(report_name):
         Get the report with the given name.
         :param str report_name: The name of the report
         :return: The JSON report as a string with a status code and string
-        :rtype: The D-Bus types (bs), q, and s
+        :rtype: The D-Bus types s, q, and s
```
mulkieran, line 408: See also uses of this method in ```stratisd_cert.py```.

jbaublitz: Can you clarify? Those have already been changed from what I can see from the diff on Github and a check in my source code.

mulkieran: Must have blinked...

PR #586, opened 2020-05-26
pool unlock command

jbaublitz:
Closes stratis-storage/project#171

### DIFF
--- a/src/stratis_cli/_actions/_top.py
+++ b/src/stratis_cli/_actions/_top.py
@@ -27,6 +27,7 @@ from tty import setcbreak
 from justbytes import Range
 
 from .._errors import (
+    StratisCliAggregateError,
     StratisCliEngineError,
     StratisCliIncoherenceError,
     StratisCliInUseOtherTierError,
@@ -789,6 +790,9 @@ class TopActions:
         """
         Unlock all of the encrypted pools that have been detected by the daemon
         but are still locked.
+        :raises StratisCliIncoherenceError:
+        :raises StratisCliNoChangeError:
+        :raises StratisCliAggregateError:
         """
         from ._data import Manager
 
@@ -798,6 +802,7 @@ class TopActions:
         if pool_uuid_list == []:
             raise StratisCliNoChangeError("unlock", "pools")
 
+        errors = []
         # This block is not covered as it would require a relatively complicated
         # simulation in the stratisd sim_engine.
         for uuid in pool_uuid_list:  # pragma: no cover
@@ -806,7 +811,7 @@ class TopActions:
             )
 
             if return_code != StratisdErrors.OK:  # pragma: no cover
-                raise StratisCliEngineError(return_code, message)
+                errors.append(StratisCliEngineError(return_code, message))
 
             if not is_some:
                 raise StratisCliIncoherenceError(
@@ -815,3 +820,6 @@ class TopActions:
                         "no new devices were unlocked during this operation"
                     )
                 )
+
+        if errors != []:
+            raise StratisCliAggregateError("unlock", "device", errors)
--- a/src/stratis_cli/_actions/_top.py
+++ b/src/stratis_cli/_actions/_top.py
@@ -811,7 +811,7 @@ class TopActions:
             if not is_some:
                 raise StratisCliIncoherenceError(
                     (
-                        "stratisd reported that some existing pools are locked but "
-                        "no new pools were unlocked during this operation"
+                        "stratisd reported that some existing devices are locked but "
+                        "no new devices were unlocked during this operation"
                     )
                 )
--- a/src/stratis_cli/_actions/_top.py
+++ b/src/stratis_cli/_actions/_top.py
@@ -146,18 +146,41 @@ def _fetch_keylist_property(proxy):
     """
     Fetch the KeyList property from stratisd.
     :param proxy: proxy to the top object in stratisd
-    :return: list of keys in the kernel keyring
+    :return: list of key descriptions
     :rtype: list of str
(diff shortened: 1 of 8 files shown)


src/stratis_cli/_actions/_top.py
```diff
@@ -760,3 +783,35 @@ def list_keys(_):
...
+        proxy = get_object(TOP_OBJECT)
+
+        pool_uuid_list = _fetch_locked_pools_property(proxy)
+        if pool_uuid_list == []:
+            raise StratisCliNoChangeError("unlock", "pools")
+
+        # This block is not covered as it would require a relatively complicated
+        # simulation in the stratisd sim_engine.
+        for uuid in pool_uuid_list:  # pragma: no cover
+            ((is_some, _), return_code, message) = Manager.Methods.UnlockPool(
```
mulkieran, line 804: I'm not sure whether the boolean in the returned value is redundant or not; I'm looking at the daemon side as well, and I'll try to return to that question shortly.

src/stratis_cli/_actions/_top.py
```diff
@@ -760,3 +783,35 @@ def list_keys(_):
...
+            )
+
+            if return_code != StratisdErrors.OK:  # pragma: no cover
+                raise StratisCliEngineError(return_code, message)
+
+            if not is_some:
+                raise StratisCliIncoherenceError(
+                    (
+                        "stratisd reported that some existing pools are locked but "
+                        "no new pools were unlocked during this operation"
```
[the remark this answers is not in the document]
jbaublitz: I was debating that wording so that sounds good.

src/stratis_cli/_actions/_top.py
jbaublitz: This follows the pattern that we use elsewhere for things like `add-data`. If you decide this is unnecesssary, we should go back and change a few other things too.

jbaublitz:
Docs are now updated.

src/stratis_cli/_actions/_top.py
mulkieran: We introduced that boolean during the idempotency work for 2.0.0 release and I feel sure that we would never have introduced a new field in the D-Bus API that we believed was entirely redundant. So, we felt that it would be possible to return either ```(false, vec![])``` indicating that nothing was done, or ```(true, vec![])``` indicating that something was done but it resulted in no action actually being taken. I'll have to take a deeper look to see what the situation is now. Adding blockdevs is a really tricky area, if we have another example of the same return structure, it might make it clearer what distinction is being made.

jbaublitz: I remember that decision and I was thinking that maybe I could change the code to follow that rule? I see what you mean that it could still be redundant because it should always be a "valid" response if the response code is 0. Does that seems like what you're seeing too?

src/stratis_cli/_actions/_top.py
```diff
@@ -760,3 +783,35 @@ def list_keys(_):
...
+
+        proxy = get_object(TOP_OBJECT)
+
+        pool_uuid_list = _fetch_locked_pools_property(proxy)
+        if pool_uuid_list == []:
+            raise StratisCliNoChangeError("unlock", "pools")
+
+        # This block is not covered as it would require a relatively complicated
+        # simulation in the stratisd sim_engine.
+        for uuid in pool_uuid_list:  # pragma: no cover
```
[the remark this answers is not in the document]
jbaublitz: I agree that storing the errors sounds more helpful to the user in case multiple pools fail to unlock for different reasons.

jbaublitz:
@mulkieran I added a new commit with `StratisCliAggregateError`. Does this look something like what you were thinking?

src/stratis_cli/_errors.py
```diff
@@ -427,3 +427,28 @@ def __str__(self):
...
+        # pylint: disable=super-init-not-called
+        self.operation = operation
+        self.type = type_
+        self.errors = errors
+
+    def __str__(self):
+        return (
+            "The operation '%s' on a resource of type %s failed. The following"
+            "errors occurred:\n%s"
+        ) % (self.operation, self.type, "\n".join(self.errors))
```
mulkieran, line 454: Please use os.linesep instead of "\n" here.

Also, you'll want to stringify each error in ```self.errors```, otherwise you'll get a ```TypeError``` on execution.

jbaublitz:
@mulkieran It could go either way but it depends on a few things:
* Would we ever want to add unlocking for one particular pool? If so, it should probably stay a pool command (something like the addition of an `--all` flag or accepting a UUID).
* Do we want to abstract away from the user that this is ultimately unlocking pool by pool? In some cases (like with keys) we've done some abstraction. In others, we're pretty straightforward about what we're doing (how we handle devices). I'm not sure if you have thoughts on which we should be doing.

jbaublitz:
@mulkieran Changes from #591 applied.

mulkieran:
> @mulkieran It could go either way but it depends on a few things:
> 
>     * Would we ever want to add unlocking for one particular pool? If so, it should probably stay a pool command (something like the addition of an `--all` flag or accepting a UUID).
> 
>     * Do we want to abstract away from the user that this is ultimately unlocking pool by pool? In some cases (like with keys) we've done some abstraction. In others, we're pretty straightforward about what we're doing (how we handle devices). I'm not sure if you have thoughts on which we should be doing.

We are really likely to be extending the command-line syntax to allow specifying a single set to unlock by UUID, which is a good reason to keep it part of the pool sub-command.

mulkieran:
@jbaublitz Please rebase.

jbaublitz:
Rebased.

PR #589, opened 2020-05-28
Make some improvements to check.py

mulkieran:
These are various improvements to check.py and its use. The last change is a convenience for those wanting to run the script on Fedora 31, and we'll revert it when we have all moved up to Fedora 31.

Related: https://github.com/stratis-storage/project/issues/175


mulkieran:
Whoops! I realized I could add a dynamic check for the pylint version, so I'll do that.

mulkieran:
I think it is really ready now.

Issue #557, opened 2020-03-14
Add additional modes to blackbox tests


mulkieran:
Currently the tests only have one mode, the daemon gets started in setup and stopped in cleanup, and actions and their correctness are verified in between. We probably need more modes:

- [ ] ~~A test of setup and device discovery~~
- [x] A test of D-Bus permission --- we want to verify that unprivileged users have restricted permissions on the D-Bus. This gets even more important with key-handling, I would say.
  - [x] Tests of the D-Bus API
  - [x] Tests of the CLI
- [x] Tests of symlinks, i.e., that they are set and then updated properly on rename of pool or filesystem. It is possible that they could be incorporated into all the existing tests via some general infrastructure. Note that these tests will not be able to be made in a sensible fashion until there is an existing spec file that install the udev rules and the script that produces the correct values for the symlink.

mulkieran:
We would like to start having some symlinks tests. How?

jbaublitz:
I feel like this could be as simple as creating a pool and filesystem in the blackbox tests and checking to see if udev created the appropriate symlink. We could even check that the symlink points to the right devicemapper device if that seems like it would be beneficial. Overall, I think we just need to verify that udev created the symlink and when we create a pool and filesystem, we already have those names available so we could simply create them both and see if a symlink exists at the expected path. How does that sound to you?

mulkieran:
@jbaublitz Yes. That seems reasonable.

mulkieran:
@bgurney: Could you try writing a symlinks test? The idea here https://github.com/stratis-storage/stratis-cli/issues/557#issuecomment-671335264 seems fine to me.

mulkieran:
Tests of setup and device discovery have the highest risk of failing for reasons outside the control of the tests or stratisd. Therefore, I'll consider this done.

PR #592, opened 2020-05-28
Add permissions check to infra.py

lleshchi:
Related #557

### DIFF
--- a/tests/blackbox/stratisd_cert.py
+++ b/tests/blackbox/stratisd_cert.py
@@ -140,10 +140,13 @@ class StratisCertify(unittest.TestCase):  # pylint: disable=too-many-public-meth
 
         self.assertEqual(return_code, expected_return_code, msg=msg)
 
-    def _test_permissions(self, dbus_method, permissions, *args):
+    def _test_permissions(self, dbus_method, args, permissions):
         """
         Test running dbus_method with and without root permissions.
-        :param dbus_method: StratisDbus wrapper method
+        :param dbus_method: D-Bus method to be tested
+        :type dbus_method: StratisDbus method
+        :param args: the arguments to be passed to the D-Bus method
+        :type args: list of objects
         :param bool permissions: True if dbus_method needs root permissions to succeed.
                                 False if dbus_method should succeed without root permissions.
         """
@@ -186,7 +189,7 @@ class StratisCertify(unittest.TestCase):  # pylint: disable=too-many-public-meth
         """
         Test that GetManagedObjects succeeds when root permissions are dropped.
         """
-        self._test_permissions(StratisDbus.get_managed_objects, False)
+        self._test_permissions(StratisDbus.get_managed_objects, [], False)
 
     def test_stratisd_version(self):
         """
@@ -198,7 +201,7 @@ class StratisCertify(unittest.TestCase):  # pylint: disable=too-many-public-meth
         """
         Test that getting daemon version succeeds when permissions are dropped.
         """
-        self._test_permissions(StratisDbus.stratisd_version, False)
+        self._test_permissions(StratisDbus.stratisd_version, [], False)
 
     def test_pool_list_empty(self):
         """
@@ -211,7 +214,7 @@ class StratisCertify(unittest.TestCase):  # pylint: disable=too-many-public-meth
         """
         Test listing pool succeeds when root permissions are dropped.
         """
-        self._test_permissions(StratisDbus.pool_list, False)
+        self._test_permissions(StratisDbus.pool_list, [], False)
 
     def test_blockdev_list(self):
         """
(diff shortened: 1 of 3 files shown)


lleshchi:
To Do: Add call to test_permissions() in stratisd_certy.py

tests/blackbox/testlib/infra.py
```diff
@@ -121,3 +125,37 @@ def __exit__(self, exception_type, exception_value, traceback):
...
+    StratisDbus.close_connection()
+
+    os.seteuid(_NON_ROOT)
+    try:
+        dbus_method(*args)
+    except dbus.exceptions.DBusException as exception:
+        if exception.get_dbus_name() == "org.freedesktop.DBus.Error.AccessDenied":
+            _PERMISSIONS = True
+        else:
+            raise exception
```
lleshchi, line 155: Is there a more efficient way to check for AccessDenied exception here?

tests/blackbox/testlib/dbus.py
```diff
@@ -466,3 +466,10 @@ def get_report(report_name):
             StratisDbus._REPORT_IFACE,
         )
         return iface.GetReport(report_name, timeout=StratisDbus._TIMEOUT)
+
+    @staticmethod
+    def close_connection():
+        """
+        Close bus connection.
+        """
+        StratisDbus._BUS.close()
```
jbaublitz, line 475: You'll need to actually reset the connection. If you just close it, any subsequent D-Bus method calls will fail. You'll need to reinitialize it too.

tests/blackbox/testlib/infra.py
```diff
@@ -121,3 +125,37 @@ def __exit__(self, exception_type, exception_value, traceback):
...
+    except Exception as exception:
+        os.seteuid(_ROOT)
+        raise exception
```
jbaublitz, line 158: It looks like you've kind of combined the approaches here. You'd probably want to drop `except Exception` or else not check the type and just match against `dbus.freedesktop.DBus.Error.AccessDenied` in the first `except` block. Currently with this code, any D-Bus error that is not access denied will leave the testing process as non-root.

tests/blackbox/testlib/dbus.py
```diff
@@ -466,3 +466,11 @@ def get_report(report_name):
...
         )
         return iface.GetReport(report_name, timeout=StratisDbus._TIMEOUT)
+
+    @staticmethod
+    def reconnect():
+        """
+        Close and reopen bus connection.
+        """
+        StratisDbus._BUS.close()
+        StratisDbus._BUS = dbus.SystemBus()
```
lleshchi, line 476: Since `StratisDbus._BUS` is reset, it is no longer a constant. Should it be renamed?

jbaublitz: If I recall correctly, it just has to do with the case of the variable (`_BUS` vs `_bus`). Try that and see if pylint is okay with that. I'm not sure how Python will react to variables being mutated in a singleton.

tests/blackbox/testlib/infra.py
```diff
@@ -121,3 +125,42 @@ def __exit__(self, exception_type, exception_value, traceback):
             if exception_value is None:
                 raise rexc
             raise rexc from exception_value
+
+
+def test_permissions(dbus_method, permissions, *args):
+    """
+    Test running dbus_method with and without root permissions.
+    param dbus_method: StratisDbus wrapper method
```
mulkieran, line 133: Missing some colons here. Try running the doc target and see what happens. Likely it'll succeed but the docs will look funky.

tests/blackbox/stratisd_cert.py
```diff
@@ -136,39 +137,107 @@ def _unittest_command(self, result, expected_return_code):
...
+    def _test_permissions(self, dbus_method, permissions, *args):
+        """
+        Test running dbus_method with and without root permissions.
+        :param dbus_method: StratisDbus wrapper method
+        :param bool permissions: True if dbus_method needs root permissions to succeed.
+                                False if dbus_method should succeed without root permissions.
+        """
+        _ROOT = 0
+        _NON_ROOT = 1
+        _permissions_flag = False
```
lleshchi, line 149: What is the best way to deal with the snake case error. Should I move the constants elsewhere, rename them, or ignore pylint?

tests/blackbox/stratisd_cert.py
```diff
@@ -136,39 +137,107 @@ def _unittest_command(self, result, expected_return_code):
 
         self.assertEqual(return_code, expected_return_code, msg=msg)
 
+    def _test_permissions(self, dbus_method, permissions, *args):
+        """
+        Test running dbus_method with and without root permissions.
+        :param dbus_method: StratisDbus wrapper method
+        :param bool permissions: True if dbus_method needs root permissions to succeed.
```
lleshchi, line 144: Do *args need to be listed here as well?

mulkieran: I'ld rather it were just ```args```, and that the arguments to the methods were passed as a list, rather than as a sequence of arguments to the method.

So,

```
:param args: the arguments to be passed to the D-Bus method
:type args: list of object
```

Example invocation:

```
self._test_permissions(
            StratisDbus.pool_add_data, True, [pool_path, StratisCertify.DISKS[2:3]]
        )
```

It's actually less confusing where this function is called.

Issue #587, opened 2020-05-27
Add TEST_RUNNER method to whitebox/integration tests and use it in test code


mulkieran:
TEST_RUNNER's job is to call RUNNER, catch any StratisCLI exception that might be raised, and then reraise the exception, chained with a subtype of AssertionError.

Since there is no handy AssertionError, our test framework should define one. We can call it something like StratisCliFailureException and it should have a message like "Unexpected failure of CLI command during test". Since this is Python, the rest should be handled by our exception mechanism.

Then ```TEST_RUNNER``` should be called in place of ```RUNNER``` wherever ```RUNNER``` is being called within a test in order to test something.

The purpose of this is to make sure that if there is a test _failure_, then the unittest mechanism identifies it as a test failure, not as a test _error_ which is what it does now.

mulkieran:
The order of steps might be:

* Define TEST_RUNNER exactly the same as RUNNER, and then make the change from RUNNER to TEST_RUNNER everywhere that is appropriate in the tests.
* Now, redefine TEST_RUNNER to be different from RUNNER
  * Define the new exception
  * Write TEST_RUNNER so it catches one exception and then re-raises the newly defined exception

bgurney-rh:
First, I have to find RUNNER.  Is this it?

```
./tests/whitebox/_misc.py:172:RUNNER = run()
```

mulkieran:
> First, I have to find RUNNER. Is this it?
> 
> ```
> ./tests/whitebox/_misc.py:172:RUNNER = run()
> ```

That's the one.

bgurney-rh:
I made a quick list of the whitebox/integration tests that run RUNNER:

```
key:
test_unset
test_list
test_reset
test_set
test_destroy

logical:
test_destroy
test_create
test_snapshot
test_list
test_rename

_misc

physical:
test_list

pool:
test_destroy
test_create
test_init_cache
test_add
test_list
test_rename
```

mulkieran:
@bgurney-rh : This code here https://github.com/bgurney-rh/stratis-cli/commit/e458059421ed7c6de8b068b9f3aebdb8beae4b0b is the right first step. Later you will want to do something like this:

```
def test_runner(command_line_args):
    try:
        RUNNER(commad_line_args)
    except StratisCliActionError as err:
        raise <the new error, not yet defined> from err

TEST_RUNNER = test_runner
```

[...]

PR #594, opened 2020-05-29
Add TEST_RUNNER for whitebox tests

bgurney-rh:
Related: #587

### DIFF
--- a/src/stratis_cli/_errors.py
+++ b/src/stratis_cli/_errors.py
@@ -399,6 +399,12 @@ class StratisCliEnvironmentError(StratisCliError):
     """
 
 
+class StratisCliTestRunError(StratisCliError):
+    """
+    Exception that occurs after a TEST_RUNNER failure.
+    """
+
+
 class StratisCliStratisdVersionError(StratisCliRuntimeError):
     """
     Raised if stratisd version does not meet CLI version requirements.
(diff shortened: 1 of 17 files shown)


bgurney-rh:
Ah, yes, isort.  I'll fix that.

src/stratis_cli/_errors.py
```diff
@@ -399,6 +399,12 @@ class StratisCliEnvironmentError(StratisCliError):
     """
 
 
+class StratisCliTestRunError(StratisCliError):
```
mulkieran, line 402: You'll want to define this class in the tests module not in the src. Probably the best place is in _misc.py, along with the definition of ```TEST_RUNNER```.

bgurney-rh: On a simple move, I'm getting a few lint errors:

- undefined-variable (for StratisCliError; should I choose something else?)
- too-few-public-methods (0/2)

bgurney-rh: I think I got it: change "StratisCliError" to "Exception".

bgurney-rh: Resolved.

tests/whitebox/_misc.py
```diff
@@ -171,6 +171,11 @@ def setUp(self):
 
 RUNNER = run()
 
+class StratisCliTestRunError(Exception):
```
mulkieran, line 175: Now make this a subtype of ```AssertionError```, that's the vital part that will the unittest framework to interpret the exception as indicating a test failure, not an error.

bgurney-rh:
I'm closing this PR, as it is now being developed against the stratis-cli 2.1.0 branch.

PR #607, opened 2020-06-15
Add blackbox tests for CLI permissions

lleshchi:
Related #557

### DIFF
--- a/tests/blackbox/stratis_cli_cert.py
+++ b/tests/blackbox/stratis_cli_cert.py
@@ -122,7 +122,7 @@ class StratisCertify(unittest.TestCase):  # pylint: disable=too-many-public-meth
         :type args: List of str
         :param bool permissions: True if the stratis command needs roots permissions to succeed,
                                     otherwise False.
-        :param bool exp_stdout_empty: True is stdout is expected to be empty
+        :param bool exp_stdout_empty: True if stdout is expected to be empty
                                         when Stratis command succeeds.
         """
         os.seteuid(_NON_ROOT)
--- a/tests/blackbox/stratis_cli_cert.py
+++ b/tests/blackbox/stratis_cli_cert.py
@@ -171,9 +171,9 @@ class StratisCertify(unittest.TestCase):  # pylint: disable=too-many-public-meth
         """
         self.unittest_command([_STRATIS_CLI, "daemon", "version"], 0, True, False)
 
-    def test_permissions_stratisd_version(self):
+    def test_stratisd_version_permissions(self):
         """
-        Test that gettign the daemon version succeeds with dropped permissions.
+        Test that getting the daemon version succeeds with dropped permissions.
         """
         self.test_permissions([_STRATIS_CLI, "daemon", "version"], False, False)
 
@@ -183,12 +183,26 @@ class StratisCertify(unittest.TestCase):  # pylint: disable=too-many-public-meth
         """
         self.unittest_command([_STRATIS_CLI, "daemon", "redundancy"], 0, True, False)
 
+    def test_stratisd_redundancy_permissions(self):
+        """
+        Test listing the redundancy levels succeeds with dropped permissions.
+        """
+        self.test_permissions([_STRATIS_CLI, "daemon", "redundancy"], False, False)
+
     def test_engine_state_report(self):
         """
         Test getting the engine_state_report.
         """
         self.unittest_command(
             [_STRATIS_CLI, "report", "engine_state_report"], 0, True, False
+        )
+
+    def test_engine_state_report_permissions(self):
+        """
+        Test getting the engine_state_report succeeds with dropped permissions.
+        """
+        self.test_permissions(
+            [_STRATIS_CLI, "report", "engine_state_report"], False, False
         )
 
     def test_invalid_report(self):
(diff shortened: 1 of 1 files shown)


tests/blackbox/stratis_cli_cert.py
```diff
@@ -113,6 +115,26 @@ def setUp(self):
         time.sleep(1)
         exec_command(["udevadm", "settle"])
 
+    def test_permissions(self, command_line, permissions, exp_stdout_empty):
+        """
+        Test running cli commands with and without root permissions.
```
waltdisgrace, line 120: Capitalize CLI

tests/blackbox/stratis_cli_cert.py
```diff
@@ -113,6 +115,26 @@ def setUp(self):
...
+        :param list args: The arguments needed to execute the Stratis command being tested
+        :type args: List of str
+        :param bool permissions: True if the stratis command needs roots permissions to succeed,
+                                    otherwise False.
```
waltdisgrace, line 124: Capitalize Stratis, typo in root

tests/blackbox/stratis_cli_cert.py
```diff
@@ -113,6 +115,26 @@ def setUp(self):
         time.sleep(1)
         exec_command(["udevadm", "settle"])
 
+    def test_permissions(self, command_line, permissions, exp_stdout_empty):
+        """
+        Test running cli commands with and without root permissions.
+        :param list args: The arguments needed to execute the Stratis command being tested
```
waltdisgrace, line 121: command_line is a parameter; args is not a parameter

tests/blackbox/stratis_cli_cert.py
```diff
@@ -113,6 +115,26 @@ def setUp(self):
...
+        :type args: List of str
```
waltdisgrace, line 122: command_line is a parameter; args is not a parameter

tests/blackbox/stratis_cli_cert.py
```diff
@@ -113,6 +115,26 @@ def setUp(self):
...
+                                        when Stratis command succeeds.
+        """
+        os.seteuid(_NON_ROOT)
+
+        if permissions:
+            self.unittest_command(command_line, 1, False, True)
+        else:
+            self.unittest_command(command_line, 0, True, exp_stdout_empty)
+
+        os.seteuid(_ROOT)
```
mulkieran, line 135: I think that, if the command should have failed because it didn't have root permissions, after the effective UID is reset to root is a good time to try to rerun the command. It should succeed in that case, or else the test is still a failure.

lleshchi: Could I get additional clarification on this? If I understand what you're saying correctly, I believe I rerun the command as root on the line below.

mulkieran: You got that right. I just somehow didn't see what I was looking at.

tests/blackbox/stratis_cli_cert.py
```diff
@@ -113,6 +115,29 @@ def setUp(self):
...
+        :param list command_line: The arguments needed to execute the Stratis command being tested
+        :type command_line: List of str
+        :param bool permissions: True if the Stratis command needs root permissions to succeed,
+                                    otherwise False.
+        :param bool exp_stdout_empty: True if stdout is expected to be empty
+                                        when Stratis command succeeds.
+        """
+
+        os.seteuid(_NON_ROOT)
+        try:
```
mulkieran, line 130: This arrangement seems fine to me, and I would expect it to behave correctly. This is so simple, actually, that there does not seem much obvious benefit to using the multiprocessing module. I feel we should go ahead with the design as is, and file an issue for the multiprocessing modification, to be pursued later.

tests/blackbox/stratis_cli_cert.py
```diff
@@ -175,12 +220,24 @@ def test_pool_list_empty(self):
         """
         self.unittest_command([_STRATIS_CLI, "pool", "list"], 0, True, False)
 
+    def test_pool_list_empty_permissions(self):
```
mulkieran, line 223: Much better for pool and filesystem to test on a non-empty list. All the interesting action happens when there is a pool, because that's when we need to use the multiiple different interfaces, FetchProperties, and whatever else, to get the various chunks of information.

tests/blackbox/stratis_cli_cert.py
```diff
@@ -113,6 +115,29 @@ def setUp(self):
...
+        try:
+            if permissions:
+                self.unittest_command(command_line, 1, False, True)
+            else:
+                self.unittest_command(command_line, 0, True, exp_stdout_empty)
+        except Exception as err:
+            os.seteuid(_ROOT)
+            raise err
+
+        self.unittest_command(command_line, 0, True, exp_stdout_empty)
```
mulkieran, line 139: I recall now that I thought this should only run if ```permissions``` is True. Otherwise, the call, whatever it is, should already have succeeded, in which case running it twice may cause a test error.

lleshchi: Ah, this makes sense, since unlike the dbus calls, the cli commands are not idempotent.

lleshchi: I also noticed that with this arrangement, the euid does not get reset to root on successful runs of the try block.

PR #609, opened 2020-06-16
Multiprocessing approach for blackbox tests

lleshchi:
This could be a potential alternative.
Related #557


tests/blackbox/stratis_cli_cert.py
```diff
@@ -115,6 +116,22 @@ def setUp(self):
         time.sleep(1)
         exec_command(["udevadm", "settle"])
 
+    def test_drop_permissions(self, command_line, permissions, exp_stdout_empty):
```
mulkieran, line 119: Prepend an ```'_'``` since this method is private. (There are other methods that should have an ```'_'``` but don't, but I'm ignoring that for now.)

PR #614, opened 2020-06-29
Distinguish between connection to D-Bus and stratisd running failure

### LINKED ISSUE #613: Distinguish cases where stratisd has actually failed to start from ones where it has failed to establish a D-Bus connection
If the file `/usr/lib/systemd/system/stratisd.service` is missing the stratisd service will start, but will not be able to bring up the dbus interface.  From a users perspective this could be confusing.

lleshchi:
Closes #613

### DIFF
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -20,6 +20,7 @@ import sys
 
 # isort: THIRDPARTY
 import dbus
+import psutil
 
 # isort: FIRSTPARTY
 from dbus_client_gen import (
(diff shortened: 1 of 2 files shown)


src/stratis_cli/_error_reporting.py
```diff
@@ -115,9 +116,18 @@ def _interpret_errors_0(error):
...
+    # can be replicated by simply not starting stratisd, the second condition
+    # is difficult to replicate through testing.
     if error.get_dbus_name() == "org.freedesktop.DBus.Error.NameHasNoOwner":
-        return "Most likely stratis is unable to connect to the stratisd D-Bus service."
+        for proc in psutil.process_iter():
+            try:
+                if proc.name() == "stratisd":  # pragma: no cover
+                    return "Most likely stratis is unable to connect to the stratis D-Bus service."
+            except psutil.NoSuchProcess:  # pragma: no cover
+                pass
```
lleshchi, line 128: Is this the best way to handle this exception?

mulkieran: Yes, I think so.

src/stratis_cli/_error_reporting.py
```diff
@@ -115,9 +116,18 @@ def _interpret_errors_0(error):
...
+
+        return "Most likely a stratisd process is not running."
```
mulkieran, line 130: You can actually be more emphatic here. Something like, "It appears that there is no stratisd process running." as your code has more or less confirmed that this is the problem.

PR #619, opened 2020-07-06
Improve error message on partial unlock failure

### LINKED ISSUE #618: When unlocking multiple pools, if some devices can't be unlocked the error message implies that no devices were unlocked
Error message
Execution failed:
stratisd reported that it did not execute every action that it would have been expected to execute as a result of the command that you requested: stratisd reported that some existing devices are locked but no new devices were unlocked during this operation 

setup 
1. create two encrypted pools
stratis key set --capture-key foo
stratis key set --capture-key moo
stratis pool create --key-desc foo p1 /dev/sdb
stratis pool create --key-desc moo p2 /dev/sdc
2. reboot
3. Set only one key
stratis key set --capture-key foo
4. Unlock pool
stratis pool unlock
 -----------
Should the error the error message say "... but no new devices were unlocked during this operation"?  this sounds like no devices were unlock, however some were unlocked.   Is there a way for us to know if some did become unlocked.  What if the devices in the last pool that were unlock, but some devices in the first pool were not unlocked?  Would stratisd still inform the user of this error?  I'll have to test this last one.

jbaublitz:
Closes #618 

@drckeefe @mulkieran I'd like to get feedback prior to code review on the new (correct) error message. There was indeed a bug in stratis-cli where if any of the unlocks failed, the entire action would halt. This has been resolved and so, as you can see, multiple unlocks can fail but execution will continue and then the aggregate failures will be communicated at the end of execution:

```
Execution failed:
An iterative command generated one or more errors: The operation 'unlock' on a resource of type device failed. The following errors occurred:
Partial action "unlock" failed for pool with UUID 1cfaf238b3fd464f969a545073bee654: Cryptsetup error: IO error occurred: Invalid argument (os error 22)
Partial action "unlock" failed for pool with UUID 65cbf7d5cc314e3d8b2eddb2a02249c1: Cryptsetup error: IO error occurred: Invalid argument (os error 22) 
```

My main question is whether we want to make changes on the stratisd side to make the message returned from the stratisd engine (currently `Cryptsetup error: IO error occurred: Invalid argument (os error 22)`) more informative. I can definitely move the issue to `project` if we decide that this message is too unclear as it stands. My personal opinion is that I would not know why it failed to unlock if I didn't write the code so it may be worth trying to make it clearer but this will be hard given our lack of type-based error chaining and the fact that there could be multiple reasons that this would fail with `EINVAL` beyond simply the key not existing. My recommendation if we do take this approach is that we specifically check if the key description exists and return a custom error as this will probably be one of those common cases where we want to be very clear why this is failing.

### DIFF
--- a/src/stratis_cli/_actions/_top.py
+++ b/src/stratis_cli/_actions/_top.py
@@ -35,6 +35,7 @@ from .._errors import (
     StratisCliNameConflictError,
     StratisCliNoChangeError,
     StratisCliPartialChangeError,
+    StratisCliPartialFailureError,
     StratisCliResourceNotFoundError,
 )
 from .._stratisd_constants import BlockDevTiers, StratisdErrors
@@ -819,14 +820,20 @@ class TopActions:
         # management of unlocked devices, so pool_uuid_list is always empty.
         errors = []  # pragma: no cover
         for uuid in pool_uuid_list:  # pragma: no cover
-            ((is_some, _), return_code, message) = Manager.Methods.UnlockPool(
-                proxy, {"pool_uuid": uuid}
-            )
+            (
+                (is_some, unlocked_pools),
+                return_code,
+                message,
+            ) = Manager.Methods.UnlockPool(proxy, {"pool_uuid": uuid})
 
             if return_code != StratisdErrors.OK:
-                errors.append(StratisCliEngineError(return_code, message))
+                errors.append(
+                    StratisCliPartialFailureError(
+                        "unlock", "pool with UUID %s" % uuid, error_message=message
+                    )
+                )
 
-            if not is_some:
+            if is_some and unlocked_pools == []:
                 raise StratisCliIncoherenceError(
                     (
                         "stratisd reported that some existing devices are locked but "
--- a/src/stratis_cli/_actions/_top.py
+++ b/src/stratis_cli/_actions/_top.py
@@ -821,7 +821,7 @@ class TopActions:
         errors = []  # pragma: no cover
         for uuid in pool_uuid_list:  # pragma: no cover
             (
-                (is_some, unlocked_pools),
+                (is_some, unlocked_devices),
                 return_code,
                 message,
             ) = Manager.Methods.UnlockPool(proxy, {"pool_uuid": uuid})
@@ -833,7 +833,7 @@ class TopActions:
                     )
                 )
 
-            if is_some and unlocked_pools == []:
+            if is_some and unlocked_devices == []:
                 raise StratisCliIncoherenceError(
                     (
                         "stratisd reported that some existing devices are locked but "
@@ -842,4 +842,4 @@ class TopActions:
                 )
 
         if errors != []:  # pragma: no cover
-            raise StratisCliAggregateError("unlock", "device", errors)
+            raise StratisCliAggregateError("unlock", "pool", errors)
--- a/src/stratis_cli/_errors.py
+++ b/src/stratis_cli/_errors.py
@@ -452,10 +452,40 @@ class StratisCliAggregateError(StratisCliRuntimeError):
 
     def __str__(self):
         return (
-            "The operation '%s' on a resource of type %s failed. The following"
+            "The operation '%s' on a resource of type %s failed. The following "
             "errors occurred:\n%s"
         ) % (
             self.operation,
             self.type,
             os.linesep.join([str(error) for error in self.errors]),
         )
+
+
+class StratisCliPartialFailureError(StratisCliRuntimeError):
+    """
+    A non-fatal error to be reported at the end as part of a StratisCliAggregateError.
+    """
+
+    def __init__(self, action, identifier_string, error_message=None):
+        """
+        Initializer.
+        :param str action: the action that failed
+        :param str identifier_string: a string that uniquely identifies the resource
+                                      for which the action failed
+        :param error_message: an optional error message for the cause of the partial
+                              failure
+        :type error_message: NoneType or str
+        """
+        # pylint: disable=super-init-not-called
+        self.action = action
+        self.identifier_string = identifier_string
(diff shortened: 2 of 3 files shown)


mulkieran:
> My recommendation if we do take this approach is that we specifically check if the key description exists and return a custom error as this will probably be one of those common cases where we want to be very clear why this is failing.

I think that this makes sense if we can do it. It's not really an error in operation, but a specific case of pre-conditions not being met, in a way that we can anticipate.

When you say "custom error", are you imagining defining a special error in the crypt module? Or are you thinking of defining one in the libcryptsetup crate?

jbaublitz:
@mulkieran Whoops! That was unclear on my part. I was thinking something along the lines of a `StratisError::Error` message that is specific to the scenario of not having a key set in the keyring. This variant seems to make the most sense because all we really want is a specific error message instructing the user on how to resolve the issue. I'll put up an issue to track that in stratisd and put this up for review after doing one last pass over it.

src/stratis_cli/_actions/_top.py
```diff
@@ -819,14 +820,20 @@ def unlock_pools(_):
         # management of unlocked devices, so pool_uuid_list is always empty.
         errors = []  # pragma: no cover
         for uuid in pool_uuid_list:  # pragma: no cover
-            ((is_some, _), return_code, message) = Manager.Methods.UnlockPool(
-                proxy, {"pool_uuid": uuid}
-            )
+            (
+                (is_some, unlocked_pools),
```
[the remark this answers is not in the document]
jbaublitz: Resolved.

jbaublitz:
Rebased. @mulkieran This should be ready for review again.

tests/blackbox/stratisd_cert.py
```diff
@@ -78,7 +78,7 @@ def make_test_filesystem(pool_path, fs_name):
     :return: Object path of the created filesystem
     """
     (
-        (filesystems_created, (array_of_tuples_with_obj_paths_and_names),),
+        (filesystems_created, (array_of_tuples_with_obj_paths_and_names)),
```
mulkieran, line 81: Please drop this edit from this PR, putting it in a separate PR is fine.

jbaublitz: That was an automatic formatting thing it looks like. Let me check my version of black.

jbaublitz: It was a black version issue. I've updated and tested with our current development issue after reverting that and it considers that successfully formatted.

drckeefe:
The error message is improvement and I think it they will be helpful.  We should consider also that the user might specifically not want to
a key(s) to control activation, which might mean that the error message could be a soft message.    

Regarding checking the key description and providing feedback to the user.   I interpretation of this would be that stratis looks up the
key description for each pool that can be unlocked, and then checks if there is a corresponding key description set in the keyring.  If
there is a pool without a corresponding key description set, then there would be a message reporting this fact.  Is that correct?

jbaublitz:
@drckeefe 

> Regarding checking the key description and providing feedback to the user. I interpretation of this would be that stratis looks up the
key description for each pool that can be unlocked, and then checks if there is a corresponding key description set in the keyring. If
there is a pool without a corresponding key description set, then there would be a message reporting this fact. Is that correct?

Yes that is the second part of clearing up the error message. Does that sound reasonable to you?

drckeefe:
This might become noisy if someone or an application (NBDE?) is using this process to unlock one pool at a time.
- set key description 
- unlock pool
- messages about keys descriptions not being set and pools not being unlocked
- then repeating for each pool 
However, the pool unlock command wasn't intended to control activation, so I think adding a message about which pools do not have an
associated key description when the pool unlock command is issued will be helpful.

mulkieran:
> This might become noisy if someone or an application (NBDE?) is using this process to unlock one pool at a time.
> 
>     * set key description
> 
>     * unlock pool
> 
>     * messages about keys descriptions not being set and pools not being unlocked
> 
>     * then repeating for each pool
>       However, the pool unlock command wasn't intended to control activation, so I think adding a message about which pools do not have an
>       associated key description when the pool unlock command is issued will be helpful.

For any sort of automatic unlocking we would try to bypass the CLI and invoke methods on the stratisd D-Bus API entirely, which would prevent this problem. On the daemon side, each pool is unlocked individually.

jbaublitz:
@drckeefe How do you feel about this error?

Here is the error with the current stratisd:
```
Execution failed:
An iterative command generated one or more errors: The operation 'unlock' on a resource of type pool failed. The following errors occurred:
Partial action "unlock" failed for pool with UUID 9df614fa7ee74ff5a010fce72b961022: Cryptsetup error: IO error occurred: Invalid argument (os error 22) 
```
and here is the error incorporating the changes in stratis-storage/stratisd#2120:
```
Execution failed:
An iterative command generated one or more errors: The operation 'unlock' on a resource of type pool failed. The following errors occurred:
Partial action "unlock" failed for pool with UUID 9df614fa7ee74ff5a010fce72b961022: Cryptsetup error: Failed with error: The key description "test1" is not currently set. Set a key with the given key description to proceed. 
```

mulkieran:
@jbaublitz: The error message is right on the money. But I'm against it including any _instructions_ about what to do next.

jbaublitz:
> @jbaublitz: The error message is right on the money. But I'm against it including any _instructions_ about what to do next.

Okay, that is unrelated to this PR, so I can change that in the stratisd PR before we merge it for sure.

jbaublitz:
@mulkieran That should be all set now. [Here's the commit fixing that](https://github.com/stratis-storage/stratisd/pull/2120/commits/ff7b3939c7ee62e00b08c6bcb81048a2db5647bf)

drckeefe:
The messages will do nicely.

PR #625, opened 2020-07-09
Add more discussion of the --capture-key and --keyfile-path options

mulkieran:


mulkieran:
I decided to give up any idea of mentioning the ```-n``` flag for ```echo```, I think the change is long enough already.

PR #630, opened 2020-08-03
Remove terminal handling from stratis-cli

jbaublitz:
Closes stratis-storage/project#209
Blocked by stratis-storage/stratisd#2163


jbaublitz:
Rebased.

jbaublitz:
@mulkieran Oh! I can add terminal handling to the sim_engine side too, but I think this PR should probably remain unchanged.

mulkieran:
@jbaublitz Yeah, we'ld better make the sim engine behave the same. 

There's a remaining minor but annoying consideration...what if stratisd is responding slowly to D-Bus requests? In that case, the user gets a chance to start typing _before_ the daemon has set the properties of stdin correctly. A possible way to make this happen might be to start a longer running filesystem operation in one window, and then quick in a hurry attempt to set the key in another window. I haven't tried, but it might work. Let's talk about this briefly in standup.

jbaublitz:
@mulkieran I'll try to take a look at the sim engine today! I agree that we should discuss this, but I think this will eventually cease to be a problem with multithreading just like with some of our other D-Bus issues.

jbaublitz:
Rebased.

mulkieran:
@jbaublitz: I think we should bump the required stratisd version in this PR as well.

jbaublitz:
@mulkieran Yes, that makes a lot of sense. I was wondering about that myself. I'll add that to the PR.

jbaublitz:
@mulkieran Ready for review again.

mulkieran (changes requested):
Ok! Please close this and open a final PR. Please keep the version bump in a separate commit.

PR #631, opened 2020-08-10
Add clean up for terminal on D-Bus error when setting keys

jbaublitz:
@waltdisgrace found an issue where the terminal does not reset itself after a failed `stratis key set --capture-key` as a normal user. This is due to the D-Bus permissions exception in this case not being handled. This does not occur with engine errors as that case is handled appropriately.

@waltdisgrace Can you test this out and see if it resolves your issue?


waltdisgrace:
I have confirmed that this resolves my issue:

```
[gchin@gchin stratis-cli]$ ./bin/stratis key set foo --capture-key
Enter desired key data followed by the return key:
Execution failed:
Most likely stratis has insufficient permissions for the action requested. 

[gchin@gchin stratis-cli]$ 
[gchin@gchin stratis-cli]$
```

PR #643, opened 2020-09-11
Update Manager interface spec to r2

mulkieran:


mulkieran:
I think this is enough...we can update for the other interfaces, which are not really necessary, when stratis-cli starts using them.

jbaublitz (approved):

PR #650, opened 2020-09-28
Verify correctness in idempotent creation of pools

### LINKED ISSUE #632: Write a blackbox test to verify correctness in idempotent creation of pools
See issue here: https://github.com/stratis-storage/stratisd/issues/2150.

Note that this is stratisd_cert.py test only; the CLI prevents creation of a pool if one has already been created entirely.

lleshchi:
Created blackbox tests of idempotent pool creation behavior.
Closes #632

### DIFF
--- a/tests/blackbox/stratisd_cert.py
+++ b/tests/blackbox/stratisd_cert.py
@@ -345,6 +345,38 @@ class StratisCertify(unittest.TestCase):  # pylint: disable=too-many-public-meth
             StratisDbus.pool_add_cache, [pool_path, StratisCertify.DISKS[2:3]], True
         )
 
+    def test_pool_create_after_cache(self):
+        """
+        Test creating existing pool after cache was added
+        """
+        pool_name = p_n()
+        pool_path = make_test_pool(pool_name, StratisCertify.DISKS[0:1])
+
+        self._unittest_command(
+            StratisDbus.pool_init_cache(pool_path, StratisCertify.DISKS[1:2]),
+            dbus.UInt16(0),
+        )
+        self._unittest_command(
+            StratisDbus.pool_create(pool_name, StratisCertify.DISKS[0:1], None),
+            dbus.UInt16(0),
+        )
+
+    def test_pool_create_with_cache(self):
+        """
+        Test creating existing pool with device already used by cache fails
+        """
+        pool_name = p_n()
+        pool_path = make_test_pool(pool_name, StratisCertify.DISKS[0:1])
+
+        self._unittest_command(
+            StratisDbus.pool_init_cache(pool_path, StratisCertify.DISKS[1:2]),
+            dbus.UInt16(0),
+        )
+        self._unittest_command(
+            StratisDbus.pool_create(pool_name, StratisCertify.DISKS[0:2], None),
+            dbus.UInt16(1),
+        )
+
     def test_pool_add_data(self):
         """
         Test adding data to a pool.


tests/blackbox/stratisd_cert.py
```diff
@@ -345,6 +345,50 @@ def test_pool_add_cache_permissions(self):
             StratisDbus.pool_add_cache, [pool_path, StratisCertify.DISKS[2:3]], True
         )
 
+    def test_pool_create_after_cache(self):
```
jbaublitz, line 348: This test is testing what we're looking for.

tests/blackbox/stratisd_cert.py
```diff
@@ -345,6 +345,50 @@ def test_pool_add_cache_permissions(self):
...
+        self._unittest_command(
+            StratisDbus.pool_init_cache(pool_path, StratisCertify.DISKS[1:2]),
+            dbus.UInt16(0),
+        )
+        self._unittest_command(
+            StratisDbus.pool_create(pool_name, StratisCertify.DISKS[0:1], None),
+            dbus.UInt16(0),
+        )
+
+    def test_pool_create_with_cache(self):
```
jbaublitz, line 364: This test is testing something different. I'm not sure if this is related.

tests/blackbox/stratisd_cert.py
```diff
@@ -345,6 +345,50 @@ def test_pool_add_cache_permissions(self):
...
+        self._unittest_command(
+            StratisDbus.pool_init_cache(pool_path, StratisCertify.DISKS[1:2]),
+            dbus.UInt16(0),
+        )
+        self._unittest_command(
+            StratisDbus.pool_create(pool_name, StratisCertify.DISKS[0:2], None),
+            dbus.UInt16(1),
+        )
+
+    def test_pool_create_cache_with_data_device(self):
```
jbaublitz, line 380: This test is testing something different. I'm not sure if this is related.

tests/blackbox/stratisd_cert.py
lleshchi: @mulkieran and I discussed adding some additional tests such as these

jbaublitz (changes requested):
The first test seems to test exactly what we're looking to test. Because of the guards I put in place in the engine, we can't really test a failure of the idempotent case so the second and third test are actually just testing that stratisd refuses to put a device in the cache and data tier at the same time which is somewhat unrelated. The first test should be sufficient for ensuring we don't introduce a regression. What are your thoughts @mulkieran?

tests/blackbox/stratisd_cert.py
mulkieran: Previously the set of existing devices was calculated as all the devices. So, if the create pool call specified the cache and data devices that were already in the existing pool the Identity constructor would be returned. This looks to me like a valid test of the change. @lleshchi One thing that you can try is to run your tests with the change in https://github.com/stratis-storage/stratisd/pull/2151 removed. If this test now fails, that indicates that the test is valuable.

tests/blackbox/stratisd_cert.py
mulkieran: I think _this_ one is unrelated. What would happen here is that the Stratis metadata would be found on the data devices that were being proposed as cache devices, and so they would be rejected. This would occur in BlockDevMgr::initialize.

mulkieran (changes requested):
I think that ```test_pool_create_cache_with_data_device``` should be dropped, but the other two look good to me.

jbaublitz (approved):

bgurney-rh (approved):

mulkieran (approved):

drckeefe (approved):

drckeefe (approved):

PR #652, opened 2020-09-28
Sync Manager interface definitions

jbaublitz:
It looks like the two Manager interface definitions got out of sync so I've edited it so that we only need to change the definition in one place.


jbaublitz:
This appears to not be an issue. Closing.

PR #658, opened 2020-10-06
Convert any relative paths of block devices into absolute paths

mulkieran:
Related: https://github.com/stratis-storage/project/issues/229


mulkieran:
I think this is sufficient, as the daemon will complain if the block device does not exist.

mulkieran:
Convert to absolute path before throwing into frozenset for correctness; if the user specified one path in two different ways then  canonicalizing the paths will map both to the same path.

jbaublitz (approved):

drckeefe (approved):

Issue #662, opened 2020-10-13
Add command to docs/stratis.txt


drckeefe:
Add text to the manpage showing the key list command. 

```
key list:
     List all Stratis key-descriptions in the kernel keyring that can be used for encryption
```

PR #663, opened 2020-10-13
Update man page: add key list command

bgurney-rh:
Closes: #662

### DIFF
--- a/docs/stratis.txt
+++ b/docs/stratis.txt
@@ -72,6 +72,8 @@ filesystem rename <pool_name> <fs_name> <new_name>::
 blockdev list [pool_name]::
 	 List all blockdevs that make up the specified pool, or all pools, if
 	 no pool name is given.
+key list::
+     List all key-descriptions in the kernel keyring that can be used for encryption.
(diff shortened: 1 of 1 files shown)


docs/stratis.txt
```diff
@@ -72,6 +72,8 @@ filesystem rename <pool_name> <fs_name> <new_name>::
 blockdev list [pool_name]::
 	 List all blockdevs that make up the specified pool, or all pools, if
 	 no pool name is given.
+key list::
+     List all Stratis key-descriptions in the kernel keyring that can be used for encryption.
```
mulkieran, line 76: I think you should actually leave out "Stratis", because we don't use that in any of the other man page entries for key operations.

mulkieran (changes requested):

docs/stratis.txt
bgurney-rh: Resolved.

mulkieran (approved):

jbaublitz (approved):

drckeefe (approved):

Issue #667, opened 2020-10-22
Extend test_filesystem_udev_symlink to verify renaming works as well


mulkieran:
Rename a filesystem and verify that the link gets moved. Then rename a pool and verify that the link gets moved.

PR #668, opened 2020-10-28
blackbox: add rename test to test_filesystem_udev_symlink

bgurney-rh:
Related: #667

### DIFF
--- a/tests/blackbox/stratisd_cert.py
+++ b/tests/blackbox/stratisd_cert.py
@@ -28,7 +28,7 @@ from tempfile import NamedTemporaryFile
 import dbus
 from testlib.dbus import StratisDbus, fs_n, p_n
 from testlib.infra import KernelKey, clean_up
-from testlib.utils import exec_command, process_exists
+from testlib.utils import exec_command, process_exists, resolve_symlink
 
 _ROOT = 0
 _NON_ROOT = 1
@@ -500,17 +500,34 @@ class StratisCertify(unittest.TestCase):  # pylint: disable=too-many-public-meth
             "/dev/mapper/stratis-1-" + pool_uuid + "-thin-fs-" + filesystem_uuid
         )
 
-        fsdevdest = os.path.abspath(
-            os.path.join(
-                os.path.dirname(filesystem_devnode), os.readlink(filesystem_devnode),
-            )
+        fsdevdest = resolve_symlink(filesystem_devnode)
+        fsdevmapperlinkdest = resolve_symlink(fs_devmapperlinkstr)
+
+        self.assertEqual(fsdevdest, fsdevmapperlinkdest)
+
+        fs_name_rename = fs_n()
+
+        self._unittest_command(
+            StratisDbus.fs_rename(fs_name, fs_name_rename), dbus.UInt16(0)
+        )
+        # Settle after rename, to allow udev to recognize the fs rename
+        exec_command(["udevadm", "settle"])
+
+        objects = StratisDbus.get_managed_objects()
+
+        pool_gmodata = objects[pool_path]
+        pool_uuid = pool_gmodata[StratisDbus.POOL_IFACE]["Uuid"]
+        filesystem_gmodata = objects[filesystem_path]
+        filesystem_uuid = filesystem_gmodata[StratisDbus.FS_IFACE]["Uuid"]
+
+        filesystem_devnode = filesystem_gmodata[StratisDbus.FS_IFACE]["Devnode"]
+
+        fs_devmapperlinkstr = (
+            "/dev/mapper/stratis-1-" + pool_uuid + "-thin-fs-" + filesystem_uuid
         )
 
-        fsdevmapperlinkdest = os.path.abspath(
-            os.path.join(
-                os.path.dirname(fs_devmapperlinkstr), os.readlink(fs_devmapperlinkstr)
-            )
-        )
+        fsdevdest = resolve_symlink(filesystem_devnode)
+        fsdevmapperlinkdest = resolve_symlink(fs_devmapperlinkstr)
 
         self.assertEqual(fsdevdest, fsdevmapperlinkdest)
 
--- a/tests/blackbox/stratisd_cert.py
+++ b/tests/blackbox/stratisd_cert.py
@@ -507,9 +507,24 @@ class StratisCertify(unittest.TestCase):  # pylint: disable=too-many-public-meth
 
         self._test_permissions(StratisDbus.fs_create, [pool_path, fs_name], True)
 
-    def test_filesystem_udev_symlink(self):
+    def test_filesystem_udev_symlink_create(self):
         """
         Test the udev symlink creation for filesystem devices.
+        """
+        pool_name = p_n()
+        pool_path = make_test_pool(pool_name, StratisCertify.DISKS[0:1])
+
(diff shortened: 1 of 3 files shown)


bgurney-rh:
This isn't quite complete, because I have this block of code that's used twice in test_filesystem_udev_symlink, and I feel like part (or all) of it could become a helper function, perhaps in StratisDbus:
```
        objects = StratisDbus.get_managed_objects()

        pool_gmodata = objects[pool_path]
        pool_uuid = pool_gmodata[StratisDbus.POOL_IFACE]["Uuid"]
        filesystem_gmodata = objects[filesystem_path]
        filesystem_uuid = filesystem_gmodata[StratisDbus.FS_IFACE]["Uuid"]

        filesystem_devnode = filesystem_gmodata[StratisDbus.FS_IFACE]["Devnode"]
  
        fs_devmapperlinkstr = (
            "/dev/mapper/stratis-1-" + pool_uuid + "-thin-fs-" + filesystem_uuid
        )

        fsdevdest = resolve_symlink(filesystem_devnode)
        fsdevmapperlinkdest = resolve_symlink(fs_devmapperlinkstr)
```

Out of the code block above, is there something that looks like it could become a good helper function?

(I could probably also use better variable names, and/or a cleanup on how the objects are accessed.)

mulkieran:
> This isn't quite complete, because I have this block of code that's used twice in test_filesystem_udev_symlink, and I feel like part (or all) of it could become a helper function, perhaps in StratisDbus:
> 
> ```
>         objects = StratisDbus.get_managed_objects()
> 
>         pool_gmodata = objects[pool_path]
>         pool_uuid = pool_gmodata[StratisDbus.POOL_IFACE]["Uuid"]
>         filesystem_gmodata = objects[filesystem_path]
>         filesystem_uuid = filesystem_gmodata[StratisDbus.FS_IFACE]["Uuid"]
> 
>         filesystem_devnode = filesystem_gmodata[StratisDbus.FS_IFACE]["Devnode"]
>   
>         fs_devmapperlinkstr = (
>             "/dev/mapper/stratis-1-" + pool_uuid + "-thin-fs-" + filesystem_uuid
>         )
> 
>         fsdevdest = resolve_symlink(filesystem_devnode)
>         fsdevmapperlinkdest = resolve_symlink(fs_devmapperlinkstr)
> ```
> 
> Out of the code block above, is there something that looks like it could become a good helper function?
> 
> (I could probably also use better variable names, and/or a cleanup on how the objects are accessed.)

You definitely should look at making this into a helper function. The way I'ld do it, at least to start,
is try to figure out what identifiers are local to the block and what aren't. If it's local to the code
block, that means it gets a value in the code block, and only that value is used, and after the end
of the code block it's not used anymore. So, ```objects``` is a good example of something that's
local. It gets a value in the assignment ```objects = StratisDbus.get_managed_objects()```, and
the value is used in the code block, but it isn't used at all after the code block. So ```objects```
would be local to the function made from that code block. ```pool_path```, though, is not local,
because the value is assigned to it before the code block is entered. So, ```pool_path``` would
be a parameter to a function made from the code block, but ```objects``` would not, it would just
be local to that function.

You might have to adjust the start or end of your code block to make the function as useful as possible.

The way to think about where the function should end is often to think about what the return value should
be. What would be the most useful thing to return from this function, if anything?

mulkieran (commented):
This looks good! Try adding the additional step of changing the pool name.

mulkieran (changes requested):
This looks great! But, I think to be consistent w/ our nearly only one assertion per test rule, we should break this test up into three tests: one test just to make the filesystem and check the symlinks, one test to make the filesystem, rename it, and test the symlinks, and one test to make the filesytem, rename the pool, and test the filesystem symlinks. I think you should be able to reuse every helper function you've already written quite well.

bgurney-rh:
OK, that makes sense.  Do we want to create four tests, for each of the four states?
- Test symlinks after creation (no renames)
- Test symlinks after creation and fs rename
- Test symlinks after creation and pool rename
- Test symlinks after creation, pool rename, and fs rename

mulkieran:
> OK, that makes sense. Do we want to create four tests, for each of the four states?
> 
>     * Test symlinks after creation (no renames)
> 
>     * Test symlinks after creation and fs rename
> 
>     * Test symlinks after creation and pool rename
> 
>     * Test symlinks after creation, pool rename, and fs rename

Yes. That seems good.

tests/blackbox/stratisd_cert.py
```diff
@@ -487,31 +517,98 @@ def test_filesystem_udev_symlink(self):
...
+        pool_name = p_n()
+        pool_path = make_test_pool(pool_name, StratisCertify.DISKS[0:1])
 
-        filesystem_devnode = filesystem_gmodata[StratisDbus.FS_IFACE]["Devnode"]
+        fs_name = fs_n()
+        filesystem_path = make_test_filesystem(pool_path, fs_name)
 
-        fs_devmapperlinkstr = (
-            "/dev/mapper/stratis-1-" + pool_uuid + "-thin-fs-" + filesystem_uuid
+        fsdevdest, fsdevmapperlinkdest = acquire_filesystem_symlink_targets(
```
mulkieran, line 535: You can drop the check here.

tests/blackbox/stratisd_cert.py
```diff
@@ -487,31 +517,98 @@ def test_filesystem_udev_symlink(self):
...
+        """
+        Test the udev symlink creation for filesystem devices after pool rename.
+        """
+        pool_name = p_n()
+        pool_path = make_test_pool(pool_name, StratisCertify.DISKS[0:1])
+
+        fs_name = fs_n()
+        filesystem_path = make_test_filesystem(pool_path, fs_name)
 
+        fsdevdest, fsdevmapperlinkdest = acquire_filesystem_symlink_targets(
```
mulkieran, line 563: And drop it here.

tests/blackbox/stratisd_cert.py
```diff
@@ -487,31 +517,98 @@ def test_filesystem_udev_symlink(self):
...
+        """
+        Test the udev symlink creation for filesystem devices after fs and pool rename.
+        """
+        pool_name = p_n()
+        pool_path = make_test_pool(pool_name, StratisCertify.DISKS[0:1])
+
+        fs_name = fs_n()
+        filesystem_path = make_test_filesystem(pool_path, fs_name)
+
+        fsdevdest, fsdevmapperlinkdest = acquire_filesystem_symlink_targets(
```
mulkieran, line 591: And here.

mulkieran (changes requested):
This looks good. I've asked you to drop certain checks, because they can be viewed as redundant. 

After that, this PR should be ready to go.

tests/blackbox/stratisd_cert.py
bgurney-rh: Resolved.

tests/blackbox/stratisd_cert.py
bgurney-rh: Resolved.

tests/blackbox/stratisd_cert.py
bgurney-rh: Resolved.

mulkieran (changes requested):
I think this is ready to go. Please close this PR and open a new one for final review.

Issue #664, opened 2020-10-19
Ensure that test names in blackbox tests contain no illegal characters


mulkieran:
Once https://github.com/stratis-storage/stratisd/pull/2276 is merged, the blackbox tests which always generate pool and filesystem names with illegal characters will start to fail. The code needs to be altered so no illegal names are generated.

PR #670, opened 2020-10-30
Prototype of workaround for D-Bus connection hang up

jbaublitz:


mulkieran:
Jenkinf failure is just coverage

jbaublitz:
Whoops! It looks like I removed a `no cover` pragma that shouldn't have been removed. Fixing now.

jbaublitz:
TODO: Strip newline from `os.read(stdin_fd)`.

jbaublitz:
@mulkieran Because we're using a pseudoterminal, I can do some ASCII control code trickery here. Simply replacing the newline with the empty string would not work because when you write that to the controller side then the controllee side of the pseudoterminal would block waiting for a newline. However, buffering read calls on a terminal stop buffering in two cases:
1. newline is reached.
2. EOF is received.

In this case, we can replace the newline with the ASCII control character that gets emitted when you hit ctrl-D. If we write this to the controller side of the pseudoterminal in stratis-cli, the terminal driver will convert that into an EOF on the stratisd side, terminating the input without a newline attached. I didn't want to have any special casing on the stratisd side, so this should work from the perspective of not adding a newline while also not allowing stratisd to ever hang.

mulkieran:
> @mulkieran Because we're using a pseudoterminal, I can do some ASCII control code trickery here. Simply replacing the newline with the empty string would not work because when you write that to the controller side then the controllee side of the pseudoterminal would block waiting for a newline. However, buffering read calls on a terminal stop buffering in two cases:
> 
>     1. newline is reached.
> 
>     2. EOF is received.
> 
> 
> In this case, we can replace the newline with the ASCII control character that gets emitted when you hit ctrl-D. If we write this to the controller side of the pseudoterminal in stratis-cli, the terminal driver will convert that into an EOF on the stratisd side, terminating the input without a newline attached. I didn't want to have any special casing on the stratisd side, so this should work from the perspective of not adding a newline while also not allowing stratisd to ever hang.

Great!

mulkieran (commented):
This looks good!

src/stratis_cli/_actions/_top.py
```diff
@@ -204,9 +206,24 @@ def _add_update_key(proxy, key_desc, capture_key, *, keyfile_path):
...
+        new_attrs = termios.tcgetattr(stdin_fd)
         print("Enter desired key data followed by the return key:")
+        new_attrs[3] = new_attrs[3] & ~termios.ECHO
+        termios.tcsetattr(stdin_fd, termios.TCSANOW, new_attrs)
+
+        # We can arbitrarily replace the newline with an ASCII control character
+        # that will cause the terminal to send EOF because readline() will only
+        # return a newline at the end of a buffered passphrase.
+        ascii_ctrl_d = chr(0x04)
+        pass_with_eof = sys.stdin.readline().replace("\n", ascii_ctrl_d)
```
mvollmer, line 221: If you anyway read the passphrase in the stratis-cli process, why bother sending it to stratisd over a pseudo terminal? Couldn't that just be a normal pipe? Or even just a string argument in the D-Bus method call?

jbaublitz: @mvollmer While a pipe is an option, sending the passphrase over the D-Bus is not. It would be visible plain text to anyone monitoring the D-Bus. With regards to why I chose a pseudoterminal, this started when I still had terminal settings being set on the stratisd side. This required that the file descriptor being passed be a terminal or pseudoterminal. Now that I've removed that restriction, the pseudoterminal and pipe should theoretically work interchangeably. Would you prefer a pipe in this case?

jbaublitz: @mvollmer I've applied the change to a pipe and it works identically as expected.

mvollmer: > While a pipe is an option, sending the passphrase over the D-Bus is not. It would be visible plain text to anyone monitoring the D-Bus. 

The system bus can only be monitored by root. Anybody else just sees the broadcasted signals. Or am I confused?

mvollmer: > I've applied the change to a pipe and it works identically as expected.

Thanks! Any reduction in complexity is a win.

src/stratis_cli/_actions/_top.py
```diff
@@ -204,20 +205,33 @@ def _add_update_key(proxy, key_desc, capture_key, *, keyfile_path):
...
+        old_attrs = termios.tcgetattr(stdin_fd)
+        new_attrs = termios.tcgetattr(stdin_fd)
         print("Enter desired key data followed by the return key:")
+        new_attrs[3] = new_attrs[3] & ~termios.ECHO
+        termios.tcsetattr(stdin_fd, termios.TCSANOW, new_attrs)
+
+        # We can arbitrarily replace the newline with the empty string
+        # because readline() will only return a newline at the end of a
+        # buffered passphrase.
+        password = sys.stdin.readline().replace("\n", "")
```
mvollmer, line 219: What about using `getpass`? https://docs.python.org/3.9/library/getpass.html

mulkieran: Ok! Please give it a try: https://github.com/stratis-storage/stratis-cli/pull/674

mulkieran:
@jbaublitz Looks like the final changes in #664 are good. Please pull down those changes and assemble a final PR.

jbaublitz:
@mulkieran I'm assuming you mean #674. I'll pull those down now.

src/stratis_cli/_actions/_top.py
mvollmer: > Ok! Please give it a try: #674

The Cockpit PoC does not use this code path, it uses --keyfile-path. So I don't think there is much value in us testing this.

Issue #661, opened 2020-10-09
Make update_introspection_data a bit more robust in the face of existing pools and filesystems


mulkieran:
Right now, if it is run twice against the sim engine w/out restarting the sim engine, it fails, because the pool has already been made AND it assumes that the result is always going to represent the pool's object path.

bgurney-rh:
I see it; when I try running `developer_tools/update_introspection_data` a second time, it outputs this traceback:
```
Traceback (most recent call last):
  File "./stratis-cli-77.77.77/developer_tools/update_introspection_data", line 172, in <module>
    main()
  File "./stratis-cli-77.77.77/developer_tools/update_introspection_data", line 139, in main
    blockdev_proxy = bus.get_object(_SERVICE, dev_object_paths[0], introspect=False)
IndexError: list index out of range
```

bgurney-rh:
First execution:
```
(Pdb) p dev_object_paths
dbus.Array([dbus.ObjectPath('/org/storage/stratis2/2')], signature=dbus.Signature('o'))
```

Subsequent executions:
```
(Pdb) p dev_object_paths
dbus.Array([], signature=dbus.Signature('o'))
```

So, how do I distinguish between those two states?

Also, is it normal and expected for the update_introspection_data program to leave a pool and filesystem in existence after it exits?  Should these be destroyed prior to exit?

mulkieran:
Leaving the pool and filesystem is fine, because we assume it is always running against the sim engine and stopping the sim engine will clear everything. (I'm thinking we should have a property on the D-Bus to indicate whether the engine is sim or real, but we don't have it yet, so that can't be checked.)

We want the script to be idempotent; it should be able to be run over and over again and produce the same output so long as there are no intervening mutating instructions sent to the sim engine.

The problem is that we have a minimalist idempotent API on the stratisd side: if the pool is already there, the return code is 0, but the pool_object_path is "/", the default and, as you see, dev_object_paths is also the default, which is the empty list, ```[]```. So, we'll have to find a way to get the pool object path, and the dev_object_paths even if the pool has already been constructed. That will require using GetManagedObjects and searching it to find the pool object path, etc. We'll want to use the dbus-python-client-gen functionality for GetManagedObjects the way that we do everywhere else in the CLI code. So, you'll need to add extra introspection data for the ObjectManager interface to the SPECS constant. You can look in _instrospect.py to find that data, happily it never changes.

mulkieran:
Closing in favor of #731.

PR #682, opened 2020-11-17
Make update_introspection_data idempotent

bgurney-rh:
Related: #661

### DIFF
--- a/developer_tools/update_introspection_data
+++ b/developer_tools/update_introspection_data
@@ -18,16 +18,30 @@ Update Stratis introspection data.
 """
 
 # isort: STDLIB
+import os
 import xml.etree.ElementTree as ET
 
 # isort: THIRDPARTY
 import dbus
+import pdb
 
 # isort: FIRSTPARTY
+from dbus_client_gen import (
+    DbusClientGenerationError,
+    managed_object_class,
+    mo_query_builder,
+)
 from dbus_python_client_gen import make_class
 
 # a minimal chunk of introspection data, enough for the methods needed.
 SPECS = {
+    "org.freedesktop.DBus.ObjectManager": """
+<interface name="org.freedesktop.DBus.ObjectManager">
+    <method name="GetManagedObjects">
+      <arg direction="out" name="objpath_interfaces_and_properties" type="a{oa{sa{sv}}}" />
+    </method>
+  </interface>
+""",
     "org.freedesktop.DBus.Introspectable": """
 <interface name="org.freedesktop.DBus.Introspectable">
 <method name="Introspect">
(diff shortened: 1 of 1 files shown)


bgurney-rh:
This is a "work in progress", but is the start of an effort to use GetManagedObjects to check if the test pool that update_introspection_data creates already exists, and set the path accordingly, to allow for subsequent runs of the tool on the same stratisd sim instance.

developer_tools/update_introspection_data
```diff
@@ -134,8 +155,12 @@ def main():
     if return_code != 0:
         exit(return_msg)
 
+    # call GetManagedObjects() here, and find the pool_object_path and dev_object_paths
```
mulkieran, line 158: This is correct. So here you want to use the ObjectManager class that you've built above.

```
managed_objects = ObjectManager.Methods.GetManagedObjects(proxy, {})
```

Then you want to use the ```pools()``` function. Something like here:
https://github.com/stratis-storage/stratis-cli/blob/develop-2.2.1/src/stratis_cli/_actions/_top.py#L310.

mulkieran (changes requested):
This looks good. I think you should carry on and add enough code so that the script gets the correct pool object path, regardless of whether or not the pool was already created. Also, pdb has modernized since the last time I used it, it has a new breakpoint() function available in Python 3.7. Take a look around and see if there's a tutorial to go through or other useful, fairly modern docs.

developer_tools/update_introspection_data
bgurney-rh: I tried this, as a test (setting a "pool_name" variable, since it's otherwise undefined right now):
```
    pool_name = "pool_name"

    managed_objects = ObjectManager.Methods.GetManagedObjects(proxy, {})
    (pool_object_path, _) = next(
        pools(props={"Name": pool_name})
        .require_unique_match(True)
        .search(managed_objects)
    )
```
...and I get this: 
```
dbus_client_gen._errors.DbusClientUnknownSearchPropertiesError: These properties in the specified query are unknown to interface "org.storage.stratis2.pool.r1": Name
```

bgurney-rh: Oh, that's right; I'm operating with "a minimal chunk of introspection data", so I need to add the new property that I'm using:
```
<property access="read" name="Name" type="s">
</property>
```

Now execution seems to continue as before (but I still need to do something.)

bgurney-rh: I need to populate dev_object_paths now.  However, I did a "before and after" comparison, to confirm that what I did above is making some progress:

Before:
```
(Pdb) pool_object_path
dbus.ObjectPath('/')
```

After:
```
(Pdb) pool_object_path
dbus.ObjectPath('/org/storage/stratis2/1')
```

developer_tools/update_introspection_data
```diff
@@ -134,8 +157,20 @@ def main():
...
+    # for pool 'pool_name'.
+    pool_name = "pool_name"
+
+    managed_objects = ObjectManager.Methods.GetManagedObjects(proxy, {})
```
mulkieran, line 164: This looks correct.

developer_tools/update_introspection_data
```diff
@@ -145,6 +180,8 @@ def main():
     if return_code != 0:
         exit(return_msg)
 
+    # call GetManagedObjects() here, and find the filesystem_object_path
```
mulkieran, line 183: And the blockdev object path as well.

mulkieran (changes requested):
Looks good so far!

PR #695, opened 2020-12-07
Print pool UUID in pool list command

mulkieran:


mulkieran:
This is trivial. If there are any objections to how the table is formatted, we can always adjust that. Here's what it looks like now.

```
Name            Total Physical   Properties                               UUID
name   512 EiB / 0 B / 512 EiB      ~Ca,~Cr   ea38a40f3b6247e28b7d6fb2c601e598
```

drckeefe (approved):

drckeefe:
Looks good to me!

Issue #700, opened 2021-01-07
Add functionality to call the EngineStateReport via the newly introduced Manager method


mulkieran:
We might want to continue to do this under the report submenu.

Currently the man pages claim that all the reports are unsupported, we could change that to start a list of supported reports.
And we would invoke that report via the dedicated D-Bus API method.

mulkieran:
Things to do:

* Update introspection data; probably best to copy it from associated PR https://github.com/stratis-storage/stratisd/pull/2371.
* Insert calls to new Manager.r4 method instead of call to the Report interface method for ```engine_state_report```. The code for this is somewhere in the actions subdirectory.
* Adjust command-line parser definition so that passing no arguments to the "report" sub-command gives the engine_state_report, i.e., make that report the default.
* Update man page to discuss supported reports and list the one report that is supported.
* Update blackbox tests as appropriate.
* Possibly update integration tests as well, but this may not be necessary.

lleshchi:
> Things to do:
> 
> * Update introspection data; probably best to copy it from associated PR [stratis-storage/stratisd#2371](https://github.com/stratis-storage/stratisd/pull/2371).
> * Insert calls to new Manager.r4 method instead of call to the Report interface method for `engine_state_report`. The code for this is somewhere in the actions subdirectory.
> * Adjust command-line parser definition so that passing no arguments to the "report" sub-command gives the engine_state_report, i.e., make that report the default.
> * Update man page to discuss supported reports and list the one report that is supported.
> * Update blackbox tests as appropriate.
> * Possibly update integration tests as well, but this may not be necessary.

@mulkieran I somehow did not see this comment and rather than alter the parser definition, I adjusted the get_report method in _top_actions to call the correct dbus method based on the report name provided. Would adjusting the parser be more optimal? [https://github.com/lleshchi/stratis-cli/tree/engine_state_report](url)

PR #701, opened 2021-01-14
Call the EngineStateReport via Manager method

lleshchi:
Closes: https://github.com/stratis-storage/stratis-cli/issues/700

### DIFF
--- a/docs/stratis.txt
+++ b/docs/stratis.txt
@@ -97,9 +97,9 @@ key unset <key_desc>::
      operations.
 report <report_name>::
         Get a report from the daemon regarding its internal state.
-        Get engine_state_report if no report name is provided. This command only
-        supports the engine_state_report. All other report names
-        and returned JSON schemas are unstable and may be changed at any time.
+        The engine_state_report name will be supported in future releases.
+        Any other report name should be considered unstable and may be removed
+        in a future release. The JSON schema of any report must always be considered unstable.
 daemon redundancy::
        List the redundancy levels that the Stratis service supports.
 daemon version::
--- a/docs/stratis.txt
+++ b/docs/stratis.txt
@@ -96,9 +96,10 @@ key unset <key_desc>::
      Unset a key in the kernel keyring so it is no longer available for encryption
      operations.
 report <report_name>::
-        Get a report from the daemon regarding its internal state. This command is
-        not supported as the report names and returned JSON schemas are unstable
-        and may be changed at any time.
+        Get a report from the daemon regarding its internal state.
+        Get engine_state_report if no report name is provided. This command only
+        supports the engine_state_report. All other report names
+        and returned JSON schemas are unstable and may be changed at any time.
 daemon redundancy::
        List the redundancy levels that the Stratis service supports.
 daemon version::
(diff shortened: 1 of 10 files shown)


lleshchi:
@mulkieran If this looks good, I will continue on with tests and man pages

mulkieran:
@lleshchi Just explicitly allow the bad-continuation errors. black is correct, pylint is wrong.

src/stratis_cli/_actions/_top.py
```diff
@@ -652,11 +652,24 @@ def get_report(namespace):
         :raises StratisCliEngineError:
         """
         # pylint: disable=import-outside-toplevel
-        from ._data import Report
+        if (
```
mulkieran, line 655: ```report_name``` will always have some non-```None``` value, due to the parser change. So it's not necessary to check for it here.

mulkieran (changes requested):
That's all I see. You should be able to carry on w/ the rest of the steps after fixing those.

jbaublitz (commented):
Other than @mulkieran's request, this looks good to me so far!

lleshchi:
@mulkieran The `get_report()` and `get_engine_state_report()` methods in testlib/dbus.py can be condensed into one method in order to satisfy the pylint error "too-many-public-methods" or error can be ignored

mulkieran:
> @mulkieran The `get_report()` and `get_engine_state_report()` methods in testlib/dbus.py can be condensed into one method in order to satisfy the pylint error "too-many-public-methods" or error can be ignored

@lleshchi: I'm fine if you disable the warning. I think the code is simpler as it is.

src/stratis_cli/_parser/_parser.py
```diff
@@ -142,7 +142,10 @@ def wrapped_func(*args):
                 (
                     "report_name",
                     dict(
-                        default=None, type=str, help=("Name of the report to display")
+                        default="engine_state_report",
+                        type=str,
+                        help=("Name of the report to display"),
+                        nargs="?",
```
mulkieran, line 148: Please add a `choices` parameter here as well with the two available names, ```engine_state_report``` and ```errored_pool_report```.

docs/stratis.txt
```diff
@@ -96,9 +96,10 @@ key unset <key_desc>::
      Unset a key in the kernel keyring so it is no longer available for encryption
      operations.
 report <report_name>::
-        Get a report from the daemon regarding its internal state. This command is
-        not supported as the report names and returned JSON schemas are unstable
-        and may be changed at any time.
+        Get a report from the daemon regarding its internal state.
+        Get engine_state_report if no report name is provided. This command only
```
mulkieran, line 100: What we want here is to indicate which names are stable and which aren't, but making a list of 1 is unappealing.

Maybe say something like:

The `engine_state_report` name will be supported in future releases. Any other report name should be considered unstable and may be removed in a future release. The JSON schema of any report must always be considered unstable.

mulkieran (changes requested):
Two final requests...otherwise it looks ready.

mulkieran (changes requested):
Looks good! Please close this and open a new PR for final review.

PR #705, opened 2021-01-28
Add flag for unhyphenated uuid format and make hypens default

lleshchi:

### DIFF
--- a/docs/stratis.txt
+++ b/docs/stratis.txt
@@ -31,6 +31,8 @@ GLOBAL OPTIONS
 	Show help on command.
 --propagate::
 	(For debugging.) Allow exceptions raised during execution to propagate.
+--unhyphenated-uuids::
+	(For listing.) Print pool and filesystem UUIDs without hyphens for list commands.
 
 COMMANDS
 --------
--- a/src/stratis_cli/_actions/_logical.py
+++ b/src/stratis_cli/_actions/_logical.py
@@ -28,7 +28,7 @@ from .._errors import (
 from .._stratisd_constants import StratisdErrors
 from ._connection import get_object
 from ._constants import TOP_OBJECT
-from ._formatting import TABLE_FAILURE_STRING, get_property, print_table
+from ._formatting import TABLE_FAILURE_STRING, get_property, print_table, to_hyphenated
 
 
 class LogicalActions:
@@ -158,6 +158,10 @@ class LogicalActions:
                 props, "Used", lambda x: str(Range(x)), TABLE_FAILURE_STRING
             )
 
+        format_uuid = (
+            (lambda mo_uuid: mo_uuid) if namespace.unhyphenated_uuids else to_hyphenated
+        )
+
         tables = [
             (
                 path_to_name[mofilesystem.Pool()],
@@ -167,7 +171,7 @@ class LogicalActions:
                 .astimezone()
                 .strftime("%b %d %Y %H:%M"),
                 mofilesystem.Devnode(),
-                mofilesystem.Uuid(),
+                format_uuid(mofilesystem.Uuid()),
             )
             for props, mofilesystem in filesystems_with_props
         ]
--- a/src/stratis_cli/_actions/_top.py
+++ b/src/stratis_cli/_actions/_top.py
@@ -40,7 +40,7 @@ from .._errors import (
 from .._stratisd_constants import CLEVIS_TANG_TRUST_URL, BlockDevTiers, StratisdErrors
 from ._connection import get_object
 from ._constants import TOP_OBJECT
-from ._formatting import TABLE_FAILURE_STRING, get_property, print_table
+from ._formatting import TABLE_FAILURE_STRING, get_property, print_table, to_hyphenated
 from ._utils import fetch_property
 
 
@@ -344,7 +344,7 @@ class TopActions:
             )
 
     @staticmethod
-    def list_pools(_):
+    def list_pools(namespace):
         """
         List all stratis pools.
         """
@@ -437,12 +437,16 @@ class TopActions:
             prop_list.append(gen_string(mopool.Encrypted(), "Cr"))
             return ",".join(prop_list)
 
+        format_uuid = (
+            (lambda mo_uuid: mo_uuid) if namespace.unhyphenated_uuids else to_hyphenated
+        )
+
         tables = [
             (
                 mopool.Name(),
                 physical_size_triple(props),
                 properties_string(mopool, props),
-                mopool.Uuid(),
+                format_uuid(mopool.Uuid()),
             )
             for props, mopool in pools_with_props
         ]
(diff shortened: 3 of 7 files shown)


src/stratis_cli/_actions/_logical.py
```diff
@@ -158,6 +161,12 @@ def filesystem_used(props):
                 props, "Used", lambda x: str(Range(x)), TABLE_FAILURE_STRING
             )
 
+        def format_uuid(mofilesystem_uuid):
```
mulkieran, line 164: I'ld just make the decision once, in the specification of the function rather than every time the function is run:

```
format_uuid = <the identity function> if namespace.unhyphenated_uuids else lambda uuid: <the transforming expression>
```

This uses ```lambda``` keyword, which is Python's way of defining anonymous functions and also its if-expression: ```<true-expr> if <bool condition> else <false-expr>```.

mulkieran (changes requested):
Just one remark about the formatting function. Please consult w/ @drckeefe about the precise command-line for the UUID format command-line option.

lleshchi:
@mulkieran If this looks good, I can add testing

mulkieran:
> @mulkieran If this looks good, I can add testing and documentation

@lleshchi Go ahead!

lleshchi:
@mulkieran Should I add more tests or is this adequate? Would it be wise to add a test somewhere that the correct uuid format is returned? If so, where is the best place to put that?

Also, "--unhyphenated-uuids" is quite long. Are we certain we'd like to keep this name?

docs/stratis.txt
```diff
@@ -31,6 +31,8 @@ GLOBAL OPTIONS
 	Show help on command.
 --propagate::
 	(For debugging.) Allow exceptions raised during execution to propagate.
+--unhyphenated-uuids::
+	(For listing.) Print pool and filesystem uuids without hyphens for list commands.
```
mulkieran, line 35: Use "UUIDs" instead of "uuids".

src/stratis_cli/_actions/_logical.py
```diff
@@ -158,6 +161,12 @@ def filesystem_used(props):
                 props, "Used", lambda x: str(Range(x)), TABLE_FAILURE_STRING
             )
 
+        format_uuid = (
+            lambda mo_uuid: mo_uuid
+            if namespace.unhyphenated_uuids
+            else str(uuid.UUID(mo_uuid))
```
mulkieran, line 167: This is actually slightly different from what I had in mind, I didn't notice before. What I want is for ```format_uuid``` to be assigned one of two functions depending on the value of ```namespace.unhyphenated_uuids```. This is just one anonymous function. For two functions, you need:

```format_uuid = lambda mo_uuid: mo_uuid if namespace.unhyphenated_uuids else lambda mo_uuid: str(uuid.UUID(mo_uuid))```

**Review by mulkieran — CHANGES_REQUESTED** [review]
Found just a couple things. Regarding ```--unhyphenated-uuids``` I think it's true that the parser will recognize the shortest disambiguating prefix. So, users will quickly learn to type ```--u```. I would also be happy w/ ```--simple-uuids```. @drckeefe Might want to way in one last time.

Regarding testing, we have no framework set up for testing the format of the output, and I think we'll always want to avoid making such a framework, as it is too much trouble. So I think the testing is quite fine as it is.

File: src/stratis_cli/_actions/_logical.py (hunk 3 above)
Comment (mulkieran) [reply src/stratis_cli/_actions/_logical.py:167]: Actually I'ld like to revise my request here:

1. Define a function that does the uuid transformation in ```_formatting.py```. Call it say, ```to_hyphenated```.
2. Then here: ```format_uuid = lambda mo_uuid: mo_uuid if namespace.unhyphenated_uuids else to_hyphenated```.

I think that's best.

**Comment by mulkieran** [comment]
@lleshchi Please rebase. We added a change to the CLI tests and to the ci scripts which will result in the output from your changes being visible in the Jenkins tests.

File: src/stratis_cli/_actions/_logical.py (hunk 3 above)
Comment (lleshchi) [reply src/stratis_cli/_actions/_logical.py:167]: @mulkieran I see, because my version still causes the function to evaluate the "if else" statement on every call to `format_uuid`, whereas with your solution, the "if else" is only evaluated once, correct? I was confused by the syntax but I see it now.

Comment (mulkieran) [reply src/stratis_cli/_actions/_logical.py:167]: @lleshchi In Python ```lambda <id> : <expr>``` is always interpreted to be a single function, even if ```<expr>``` has ```lambda``` in it.

```
>>> thing = True
>>> func = lambda x: x if thing else lambda x: x + 1
>>> func(1)
1
>>> thing = False
>>> func(1)
<function <lambda>.<locals>.<lambda> at 0x7f8ef62948b0>
```

Not what was intended. To persuade Python to parse properly, we need some parens:

```
>>> thing = False
>>> new_func = (lambda x: x) if thing else (lambda x: x + 1)
>>> new_func(1)
2
```
Again:

```
>>> thing = True
>>> new_func = (lambda x: x) if thing else (lambda x: x + 1)
>>> new_func(1)
1
```

Note that the value of ```new_func``` is either ```lambda x: x``` or ```lambda x: x + 1``` once the assignment expression is evaluated.

Note that this means that, even with the separate function defined in ```_formatting.py``` you must still use the parens to override Python's default parsing for lambda-expressions.

File: src/stratis_cli/_actions/_top.py [hunk 4]
```diff
@@ -926,19 +930,20 @@ def unlock_pools(namespace):
         # This block is not covered as the sim engine does not simulate the
         # management of unlocked devices, so locked_pools is always empty.
         errors = []  # pragma: no cover
-        for uuid in locked_pools:  # pragma: no cover
+        for pool_uuid in locked_pools:  # pragma: no cover
```
mulkieran, line 933: These edits on ```unlock_pools``` all look like stray edits, so I think you can remove them.

src/stratis_cli/_parser/_logical.py
```diff
@@ -55,7 +55,7 @@
                 (
                     "pool_name",
                     dict(action="store", default=None, nargs="?", help="Pool name"),
-                )
+                ),
```
mulkieran, line 58: This looks like a stray edit, although I'm surprised that black doesn't either enforce or prohibit. Please see if you can remove it.

src/stratis_cli/_parser/_parser.py
```diff
@@ -164,7 +164,11 @@ def wrapped_func(*args):
 ]
 
 GEN_ARGS = [
-    ("--propagate", dict(action="store_true", help="Allow exceptions to propagate"))
+    ("--propagate", dict(action="store_true", help="Allow exceptions to propagate")),
+    (
+        "--unhyphenated-uuids",
+        dict(action="store_true", help=("List unhyphenated format of UUIDs")),
```
mulkieran, line 170: "Display UUIDs in unhyphenated format" is probably a bit better.

mulkieran (changes requested):
A few minor edits requested, otherwise I think it's ready.

mulkieran (approved):
This looks good to me.

drckeefe (approved):

jbaublitz (approved):

PR #712, opened 2021-03-03
github actions: pin Ubuntu version to 18.04

bgurney-rh:


drckeefe (approved):

mulkieran (approved):

mulkieran:
It's not actually necessary not to break the tests in this case...but it makes sense to be consistent, so might as well.

PR #713, opened 2021-03-05
Handle an exception chain connected with a GetManagedObjects() failure

mulkieran:
Related: https://github.com/stratis-storage/stratisd/pull/2404

### DIFF
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -28,7 +28,7 @@ from dbus_client_gen import (
     DbusClientMissingSearchPropertiesError,
     DbusClientUniqueResultError,
 )
-from dbus_python_client_gen import DPClientInvocationError
+from dbus_python_client_gen import DPClientInvocationError, DPClientMethodCallContext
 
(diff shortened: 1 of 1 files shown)


mulkieran:
@bgurney-rh Can you test this using the scripts you were running yesterday? We agree that the errors the CLI was encountering were Ok. What we don't like is the stack traces, since we've defined any stack trace appearing under normal operation as a bug in the CLI. So, in lieu of the stack trace, we want to see the error message added in this commit.

jbaublitz (approved):

mulkieran:
ok, testing works out.

mulkieran:
Looks like this:

```
> stratis pool list
Execution failed:
A D-Bus method failed during execution of the selected command. Most likely, the failure was due to a transient inconsistency in the D-Bus interface and the command will succeed if run again. 
```

instead of the previous stack trace.

mulkieran:
rebased.

mulkieran:
we can ignore the prior, bogus test failure, probably network related.

drckeefe (approved):

PR #718, opened 2021-03-08
Update to use r4 revision of the FetchProperties interface

mulkieran:


mulkieran:
Expect default test to fail, all others should succeed.

mulkieran:
default test should succeed now that https://github.com/stratis-storage/stratisd/pull/2431 is merged.

PR #729, opened 2021-03-22
Update filesystem Devnode property so that it is invalidates

mulkieran:
Related https://github.com/stratis-storage/stratisd/issues/2219


mulkieran:
Forgot to update this little bit after merging multi-threading PR.

PR #734, opened 2021-03-22
Remove dead key_description parameter from binding

### LINKED ISSUE #733: Investigate removing key description as a parameter for binding to Clevis
We may not need to pass in the key description as a parameter to the Clevis bind command because it's stored in our metadata.

jbaublitz:
Closes #733


mulkieran (changes requested):
Looks good, but better do the man pages as well.

mulkieran (changes requested):
Looks good to me. Please close this and open new PR for final review.

PR #737, opened 2021-03-24
Some improvements to update_introspection_data

### LINKED ISSUE #731: Automate selection of interfaces in update_introspection_data
Right now there's a chunk of code that the user has to edit in order to choose the correct interfaces. It's marked with EDIT. We want to let our script choose the correct interfaces by reading the introspection data and choosing the most recent revision.

Ideally, this script would then be runnable in CI, because the user does not need to do any editing.

We should also probably pick up the idempotency problem when we do this, as edits will definitely intersect.

mulkieran:
Closes #731 

* In update_introspection_data calculates revision numbers from stratisd version number.
* Lookup required interface names in introspection data and not vice-versa.
* Update_interface.py to be consistent w/ current stratisd version
* Calculate revision numbers from required minimum stratisd minor version in stratis CLI

### DIFF
--- a/developer_tools/update_introspection_data
+++ b/developer_tools/update_introspection_data
@@ -100,10 +100,14 @@ def _add_data(proxy, interfaces, table):
     """
     string_data = Introspectable.Methods.Introspect(proxy, {})
     xml_data = ET.fromstring(string_data)
-    for interface in xml_data:
-        interface_name = interface.attrib["name"]
-        if interface_name in interfaces:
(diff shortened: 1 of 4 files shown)


mulkieran:
Some improvements I'm deferring until later:

* Handle the idempotency problem.
* Incorporate the functionality, not necessarily in the same format,  into some sort of CI in stratisd repo. It would do two things:
  1. Verify that the revision numbers in the stratisd API always match its announced version minor number.
  2. Verify that the test interfaces and what not are properly updated.

jbaublitz (approved):

Issue #660, opened 2020-10-09
Test that our device paths are always absolute using mock objects


mulkieran:
We can do this exclusively in the CLI.

mulkieran:
We'ld have to synthesize a valid relative path from the current working directory and the known absolute path. We can do that, with ```os.path.relpath```.

Why do it in the blackbox instead of the whitebox tests? Because the sim engine doesn't care if paths are real or not. But we could do a fake test which invoked the methods, and used a mock object to determine that the arguments were correct.

Right now this all seems like too much trouble.

mulkieran:
Actually, monkey patching and using mock objects really doesn't seem all that bad. And we may someday have to use mock objects for something harder to check, so this might be a good start.

PR #744, opened 2021-04-06
Test absolute path passed to create pool method

lleshchi:
Closes: #660

### DIFF
--- a/tests/whitebox/monkey_patching/test_absolute_path.py
+++ b/tests/whitebox/monkey_patching/test_absolute_path.py
@@ -1,0 +1,105 @@
+# Copyright 2021 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Test handing of relative paths
+"""
+
+# isort: STDLIB
+import os
+
+# isort: LOCAL
+import stratis_cli
+
+from .._misc import SimTestCase, RUNNER, TEST_RUNNER
+from stratis_cli._stratisd_constants import StratisdErrors
+
+
+class RelativePathCreatePool(SimTestCase):
+    """
+    Test that relative path is converted to absolute
+    """
+
+    def test_create_pool_relative_path(self):
+        """
+        Verify that create pool receives absolute path
+        """
+        
+        def absolute_path_check(_, args):
+            self.assertTrue(all(os.path.isabs(path) for path in list(args["devices"])))
+            return ((True, (_, _)), StratisdErrors.OK, "")
+   
+        # pylint: disable=import-outside-toplevel
+        from stratis_cli._actions import _data
+
+        # pylint: disable=protected-access
+        stratis_cli._actions._data.Manager.Methods.CreatePool = absolute_path_check
+
+        command_line = ["--propagate", "pool", "create", "test_pool", "./fake/relative/path", "~/very/fake/path", "/fake/path"]
+        RUNNER(command_line)
+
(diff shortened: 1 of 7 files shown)


lleshchi:
@mulkieran If this looks good so far, I can write similar tests for other commands which use device paths. What do you think of the return of the `CreatePool` mock object?

tests/whitebox/monkey_patching/test_absolute_path.py
```diff
@@ -0,0 +1,49 @@
...
+            self.assertEqual(list(args["devices"])[0], os.path.abspath("/fake/path"))
+            return ((True, (_, _)), StratisdErrors.OK, "")
+
+        # pylint: disable=import-outside-toplevel
+        from stratis_cli._actions import _data
+
+        # pylint: disable=protected-access
+        stratis_cli._actions._data.Manager.Methods.CreatePool = absolute_path_check
+
+        command_line = ["--propagate", "pool", "create", "test_pool", "/fake/path"]
```
mulkieran, line 48: To cover all, you should have one test with an absolute path and one with a relative path.

tests/whitebox/monkey_patching/test_absolute_path.py
```diff
@@ -0,0 +1,49 @@
...
+    Test that relative path is converted to absolute
+    """
+
+    def test_create_pool_relative_path(self):
+        """
+        Verify that create pool recieves absolute path when relative path was given in command line
+        """
+
+        def absolute_path_check(_, args):
+            self.assertEqual(list(args["devices"])[0], os.path.abspath("/fake/path"))
```
mulkieran, line 39: I'ld rather here that you combine ```assertTrue``` and ```isabs```, rather than ```assertEqual``` and ```abspath```.

Something like: ```self.assertTrue(all(list(args["devices"]), os.path.isabs)``` very roughly, as that's probably not valid Python. The point is that we can use the ```all``` to pass all the device arguments somehow and use os.path.isabs to verify that each and every one is an absolute path.

mulkieran (changes requested):
Just a few requests...

mulkieran:
This all looks good; we decided last time to put each absolute path test case in a separate file, for simplicity's sake.

lleshchi:
> This all looks good; we decided last time to put each absolute path test case in a separate file, for simplicity's sake.

@mulkieran Yes, just added those changes. The only other revision I'd like to add is creating a `_misc.py` file to hold the `absolute_path_check` function that is shared between `test_abspath_init_cache.py` and `test_abspath_add.py`

mulkieran:
> > This all looks good; we decided last time to put each absolute path test case in a separate file, for simplicity's sake.
> 
> @mulkieran Yes, just added those changes. The only other revision I'd like to add is creating a `_misc.py` file to hold the `absolute_path_check` function that is shared between `test_abspath_init_cache.py` and `test_abspath_add.py`

Ok! You may have problems with the CI or pylint for that matter because of your path (PYTHONPATH specifically). If it ultimately proves less trouble to leave the duplicates in, that's also probably acceptable.

tests/whitebox/monkey_patching/test_abspath_add.py
```diff
@@ -0,0 +1,90 @@
...
+    Test that relative path is converted to absolute
+    """
+
+    _POOLNAME = "mypool"
+
+    def absolute_path_check(self, _, args):
+        """
+        Assert that all device paths passed received by method are absolute
+        """
+        self.assertTrue(all(os.path.isabs(path) for path in list(args["devices"])))
```
mulkieran, line 39: I think ```args["devices"]``` is a list already, so listing it again is extra work.

mulkieran (changes requested):

tests/whitebox/monkey_patching/test_abspath_add.py
lleshchi: @mulkieran Ah, yes you're right. In initial testing I was accessing individual items in the frozenset, but  now that I'm iterating, the list is no longer necessary

lleshchi:
@mulkieran The last thing to consider is whether to have `RelativePathCreatePool` inherit from `PathTestCase` and make use of `check_absolute_path()`. This is tricky because `stratis_cli._actions._data.Manager.Methods.CreatePool` has a different return type from the other mocked methods. It might be simpler to keep the current structure.

jbaublitz (approved):

mulkieran:
Superceded by #750

PR #755, opened 2021-05-05
Expose physical path if different from metadata path

### LINKED ISSUE #754: Display the base device as well as the encrypted device in blockdev list when the device is encrypted
Something like ```/dev/dm-4 (on /dev/sdb)```. This will be in two parts:
1. ~~Expose the information over the D-Bus. Should this be in a separate property, or in the same, multi-part property.~~ PhysicalPath property is already exposed over the D-Bus. So this is purely a CLI issue.
2. Adapt the CLI to use the information.
~~We can drop in the subissues later.~~

mulkieran:
Closes #754

### DIFF
--- a/docs/stratis.txt
+++ b/docs/stratis.txt
@@ -184,9 +184,14 @@ FIELDS for stratis blockdev list
 Pool Name::
 	  The name of the pool using the block device.
 Device Node::
-	  The device node of the block device.
+	  The device node of the block device. A second device node will
+          be displayed in parentheses if the block device is encrypted. This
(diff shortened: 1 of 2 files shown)


mulkieran:
Test failure is spurious...

drckeefe (approved):

mulkieran:
@drckeefe Having verified that the format is acceptable to everybody, I've updated the man pages and re-requested your review. As far as I am concerned, this is ready to go.

drckeefe (approved):

PR #756, opened 2021-05-07
Make a simple script to call GetManagedObjects and outputs as JSON

mulkieran:


mulkieran:
Turns out the code wasn't actually linting ```update_introspection_data``` at all, but I'll fix that in a separate commit. ```get_managed_objects``` is passing the linter.

PR #769, opened 2021-05-18
Demos

mulkieran:
get_pool_uuid_1 uses Stratis libraries for D-Bus
get_pool_uuid_2 uses dbus-python library

Stratis libraries use dbus-python transitively, they just auto-generate
a bunch of code.


mulkieran:
This was just to illustrate working w/ the D-Bus using two different infrastructures, obsolete.

PR #770, opened 2021-05-19
Add nightly job for next Fedora python check

bgurney-rh:
Supersedes: #758 
Related: stratis-storage/project#294

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,45 @@
+---
+name: stratis-cli nightly
+
+# yamllint disable-line rule:truthy
+on:
+  schedule:
+    - cron: 25 3 * * *
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  # Nightly to show us in advance what will change in Python utilities in
+  # the next version of Fedora
+  next-fedora-python-checks:
+    continue-on-error: true
+    strategy:
+      matrix:
+        include:
+          # PYTHON CHECKS ON NEXT FEDORA PYTHON AND PYTHON TOOLS VERSION
+          - dependencies: >
+              pylint
+              python3-dateutil
+              python3-dbus-client-gen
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,51 @@
...
+              python3-dbus-client-gen
+              python3-dbus-python-client-gen
+              python3-psutil
+              python3-pyudev
+              python3-requests
+              python3-semantic_version
+              python3-pygithub
+            image: fedora:33
+            task: >
+              (cd code_maintenance; make -f Makefile lint) &&
```
mulkieran, line 31: These are directories that mostly don't exist in stratis-cli. pylint checks should be much simpler on stratis-cli because stratis-cli is all Python. See the Makefile and main.yml for examples. The same holds true for formatting task.

mulkieran (changes requested):

.github/workflows/nightly.yml
bgurney-rh: It looks like a lot of things changed since I staged the last PR.  I see that main.yml primarily runs `task: PYTHONPATH=./src make -f Makefile lint` and `task: make -f Makefile fmt-ci`.

Also, on F33 right now, `make -f Makefile fmt-ci` has some failures.

mulkieran: @bgurney-rh Yeah. Starting from scratch is what you want to be doing, e.g., don't include dependencies, like pygithub, which aren't used in this repo. Throw in what makes sense, using ```main.yml``` to figure out what dependencies to use and what Makefile targets to call. Then, once the first effort is in, we can merge it, and run it. Then we can start fixing it up as necessary.

bgurney-rh: I just added a commit; does it address most of what we need?

bgurney-rh:
Rebased all commits into a single commit.

mulkieran (approved):

PR #722, opened 2021-03-09
Mulkieran develop 2.3.0 fedora 33 updated

bgurney-rh:
Related: stratis-storage/project#263

PR #774, opened 2021-05-21
Bump Fedora versions in CI

bgurney-rh:
Related: stratis-storage/project#263
Supersedes: #722

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,14 +30,14 @@ jobs:
               python3-psutil
               python3-semantic_version
               python3-wcwidth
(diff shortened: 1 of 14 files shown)


bgurney-rh:
I think we ran into the "invalid-str-returned" lint before; I don't remember if we need to lint that out, or remediate it, so I made that into a separate commit, in case I need to leave it behind.

bgurney-rh:
I can see that in #722, we had decided to disable invalid-str-returned, so it looks like I'm heading in the right direction.

mulkieran (approved):

PR #809, opened 2021-07-13
Use LockedPoolsWithDevs property for unlock method

mulkieran:


mulkieran:
All tests should succeed.

PR #820, opened 2021-08-04
Reintroduce psutil-based error message construction

mulkieran:


mulkieran:
@drckeefe This should satisfy you, since the import is now optional. I can cut another release into Fedora with this change, also.

mulkieran:
Tested with:
  1. python3-psutil installed and without
  2. stratisd running and not

drckeefe (approved):

PR #811, opened 2021-07-15
Issue project 299

mulkieran:
Related https://github.com/stratis-storage/project/issues/299

PR #822, opened 2021-08-12
Add size option for fs create command

mulkieran:
Supercedes #811


mulkieran:
All actual tests will fail.

mulkieran:
tests should pass now.

PR #825, opened 2021-08-24
Use packaging instead of semantic_version

### LINKED ISSUE #803: PendingDeprecationWarning because of spec class in semantic_version
We would prefer not to emit it, but we seem to be in some circumstances.

```
stratis-cli/src/stratis_cli/_actions/_stratisd_version.py:35: PendingDeprecationWarning: The Spec() class will be removed in 3.1; use SimpleSpec() instead.
```

The problem is that we must run with a somewhat wide range of the semantic_version Python package: 2.6.0 - 2.8.4 inclusive. ```SimpleSpec``` was not introduced until 2.7, but the PendingDeprecationWarning was added sometime before 2.8.4.

mulkieran:
Closes #803


mulkieran:
I'm very willing to extend the PR to make this version check unconditional again. The version check is run once for any invocation of stratis. This means one additional, relatively cheap, D-Bus property access, which seems like not a big deal to me.

jbaublitz (approved):

bgurney-rh (approved):

drckeefe (approved):

PR #844, opened 2021-11-16
Advance next development environment to Fedora 35

bgurney-rh:
Related: stratis-storage/project#363


mulkieran:
This'll need to be retargeted, since I released stratis-cli last night. Otherwise, good.

PR #852, opened 2022-01-22
Improve handling of options in the bash completion

carzacc:
Hi everyone,

I'm back maintaining this Bash completion script. There hasn't been a whole lot of changes to the CLI interface (that I've been able to see, anyway, let me know if there's something significant missing), so this PR mostly addresses the fact that the current script doesn't handle the possibility of having options before the word that is currently being completed. This PR should address most common cases (for example using `--propagate` before the root subcommand, as per the man page), but I will do some more digging and testing over the next few days.

Meanwhile, please give me feedback about its current state (which seems to be working, from my initial testing).

### DIFF
--- a/shell-completion/bash/stratis
+++ b/shell-completion/bash/stratis
@@ -1,6 +1,6 @@
 _stratis()
 {
-	local cur prev x2prev opts root_subcommands pool_subcommands fs_subcommands blockdev_subcommands daemon_subcommands complete_pools blockdevs first second third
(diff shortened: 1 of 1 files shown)


bgurney-rh:
Hi @carzacc,

This seems to be based on stratis-cli 2.4.0.  Can you create a version of this that is based on stratis-cli 3.0.0?

carzacc:
> Hi @carzacc,
> 
> This seems to be based on stratis-cli 2.4.0. Can you create a version of this that is based on stratis-cli 3.0.0?

You're right, I've made a bit of a mess with the branches it seems, I'll open a new PR then

PR #853, opened 2022-01-26
improved handling of options in the Bash completion

carzacc:
This time the branches look right

### DIFF
--- a/shell-completion/bash/stratis
+++ b/shell-completion/bash/stratis
@@ -1,18 +1,13 @@
 _stratis()
 {
-	local cur prev x2prev opts root_subcommands pool_subcommands fs_subcommands blockdev_subcommands daemon_subcommands complete_pools blockdevs first second third
+	local comp_cword_adj root_subcommand subcommand subcommand_opt cur prev opts root_subcommands pool_subcommands fs_subcommands blockdev_subcommands daemon_subcommands complete_pools blockdevs
 
 	COMPREPLY=()
 
 	cur="${COMP_WORDS[COMP_CWORD]}"
 	prev="${COMP_WORDS[COMP_CWORD-1]}"
-	x2prev="${COMP_WORDS[COMP_CWORD-2]}"
-	first="${COMP_WORDS[1]}"
-	second="${COMP_WORDS[2]}"
-	third="${COMP_WORDS[3]}"
 	opts="-h --help"
-	global_opts="--version --propagate ${opts}"
-	listing_opts="--unhyphenated-uuids"
+	global_opts="--version --propagate ${opts} --unhyphenated-uuids"
 	root_subcommands="daemon pool blockdev filesystem key report"
 	pool_subcommands="add-cache add-data bind create destroy init-cache list rename unbind unlock"
 	pool_create_opts="--redundancy --key-desc"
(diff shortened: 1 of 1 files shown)


carzacc:
I still want to address the references I make to the third word, that the script assumes is the place where it can find the first option for the subcommand: in this new and more flexible script it just doesn't cut it anymore to assume that, it made sense to me at the time as a way to introduce flexibility but it's not enough really.

carzacc:
Now I would be quite happy to merge this PR already, just leaving a note: for commands like filesystem destroy where the user can pick more than one value from a list I would like not to suggest to the user again stuff they've already picked, but I think we can leave that to a future PR.

bgurney-rh:
I would be careful about auto-completing items after a "destroy" sub-command, to ensure that the user is making a conscious decision to destroy something they intend to destroy (hopefully after they have confirmed, via the output of "stratis pool list" or "stratis fs list", the item they wish to destroy).

carzacc:
> I would be careful about auto-completing items after a "destroy" sub-command, to ensure that the user is making a conscious decision to destroy something they intend to destroy (hopefully after they have confirmed, via the output of "stratis pool list" or "stratis fs list", the item they wish to destroy).

The issue I saw (which is the one I mentioned in my comment) is that the user will receive autocomplete suggestions for filesystem "filesystem-one" again when they type `stratis fs destroy pool-name filesystem-one fil` and they hit Tab. 

To me, it seems you're suggesting the script shouldn't autocomplete filesystem names at all as a way to encourage user to use commands like `fs list` and do more thorough checking before issuing `fs destroy`.

It would be a fair point, and I agree that autocomplete makes it more likely that users will make this kind of choice lightly.

I don't think, though, that removing features from autocomplete, and making it less convenient to use (for example because of having to type long names) for everyone, just to encourage users to perform more thorough checks and know the exact filesystem name before issuing destructive commands is the best choice. Another thing to consider is that, to delete filesystems, there are two choices to make: first the user has to choose a pool and then a filesystem from that pool, which already narrows down the choice in some cases.

An alternative might be to add a prompt to the CLI asking users to confirm destructive action, perhaps including other data present in `fs list` like the creation date, the path and size which might cause a user to doubt what they're doing.

To avoid any potential confusion, I'm just trying contribute my own personal view to this discussion, if you tell me to remove all autocompletion for `destroy` subcommands I'll do it ASAP.

I hope my phrasing was clear.

bgurney-rh:
Actually, after testing it, I think it's okay as is, and I was probably just being a little too cautious.

After some preliminary testing, the auto-complete helped me find the right sequence to set up an encrypted pool, after having known the general steps to do so, but forgetting the exact steps.  I think that's a pretty good result, considering that completion is supposed to help the user who is in this current state.

bgurney-rh:
There's a few occurrences of trailing spaces that I'll highlight in a few minutes (including an extra line at the end of `shell-completion/bash/stratis` that was added).

shell-completion/bash/stratis
```diff
@@ -1,70 +1,116 @@
...
+				fi
+			else
+				if [[ ${root_subcommand} != "" && ${subcommand} != "" && ${subcommand_opt} == "" ]]; then
+					subcommand_opt=${el}
+				fi
+				let "comp_cword_adj--"
+			fi
+
+		done
+		# find the subcommand
```
bgurney-rh, line 51: There is a space at the end of `# find the subcommand`, and there are 4 tabs on the empty line on line 55, just before `break`.

shell-completion/bash/stratis
```diff
@@ -1,70 +1,116 @@
...
+		case ${root_subcommand} in
+			pool)
+				if [[ ${subcommand} == "create" ]]; then
+					COMPREPLY=( $(compgen -W "${pool_create_opts} ${opts}" -- ${cur}) )
+				elif [[ ${subcommand} == "bind" ]] && [[ ${subcommand_first_param} != "tpm2" ]]; then
+					COMPREPLY=( $(compgen -W "${pool_bind_opts} ${opts}" -- ${cur}) )
+				else
 					COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
+				fi
+
```
bgurney-rh, line 85: There are four tabs on the empty line of line 85

shell-completion/bash/stratis
```diff
@@ -1,70 +1,116 @@
...
 	daemon_subcommands="redundancy version"
 	key_subcommands="set unset reset list"
 	key_set_reset_opts="--keyfile-path --capture-key"
 	blockdevs="$(echo /dev/disk/*/*) $(lsblk -n -o name -p -l)"
+	root_subcommand=""
+	subcommand=""
+	subcommand_first_param=""
+	subcommand_opt=""
+	comp_cword_adj=${COMP_CWORD}
+
```
bgurney-rh, line 29: There is one tab on the empty line on line 29.

carzacc:
> There's a few occurrences of trailing spaces that I'll highlight in a few minutes (including an extra line at the end of `shell-completion/bash/stratis` that was added).

Done! If it looks good, I can squash the commits.

bgurney-rh:
I think it looks good; go ahead and squash.

mulkieran:
@carzacc Plz rebase and I'll merge this PR.

carzacc:
> @carzacc Plz rebase and I'll merge this PR.

Is it good to merge now?

mulkieran:
Yes, thx.

Issue #854, opened 2022-01-29
Add support for using shfmt to format tab-completion scripts


mulkieran:
We'll need Makefile targets and some kind of task in the CI. I think it is better if the tasks are kept separate from our other formatting tasks. Our contributors who contribute to the tab-completion scripts don't necessarily have our Python formatter installed. stratisd already has a good example.

PR #855, opened 2022-02-04
Makefile: add fmt-shell targets for bash completion

bgurney-rh:
Closes: #854


mulkieran:
@bgurney-rh Looks good. I think you should run the new makefile target as another task in the python-checks job in main.yml and also in next-fedora-python-checks job in weekly.yml.

PR #858, opened 2022-02-25
Stratisd issue 2814

mulkieran:


mulkieran:
Failing test because too low fs-limit results in D-Bus error, not engine error.

mulkieran:
rebased

mulkieran:
@bgurney-rh, @drckeefe, @jbaublitz Here's a proposed way of listing pool information. It right now comes to 5 separate values, which I think is too much. There are three size/usage values and two "free" values. I think that, to many, seeing the free values is actually more important.

So, I propose we drop the middle value, which is a used value. That brings us down to four, which is probably not too many, and I think these four are the easiest to explain.

There is one problem though. My explanation of the last value is not quite correct, because it does not take into account the Stratis metadata on the devices. If we caused the total physical size value to omit the Stratis metadata, then everything would be correct, and I don't think that it would be less easy to follow. The only problem w/ that proposal is that there might seem to be an apparent inconsistency between the pool listing and the blockdev listings, which do show the total size of the devices.

mulkieran:
> @mulkieran I'm going to defer to @drckeefe on which values to display. However I did want to ask why you're saying allocated size doesn't include space allocated for metadata. After looking through the code, I'm pretty confident that it is including both the Stratis block device metadata and the allocated space for both thin pool data and metadata. Can you point to where you're getting the information that it doesn't include that? If it isn't including that, shouldn't we change it to make sure that it does?

@jbaublitz When I look at ```DataTier::new()``` method it creates a DataTier w/ an empty vec for the ```segments``` field, but a backstore where each blockdev's metadata have already been set aside. The total allocated size is derived ultimately from DataTier::allocated() which sums up the values in the ```segments``` field. So, I'm convinced that that value does not include the space set aside for the Stratis metadata.

drckeefe:
In the output of all 5 stats a user might see (this is fictional)

Pool Name  Total Size  Allocated Size  Allocated Used  Allocated Free  Total Free
  Pool 1         3TiB             384GiB            287GiB               106GiB            2.6TiB

This amount of information is good for a detailed view.   

For a basic view I could see this: 
- The Total Size - the total amount of space in available to the pool.  Or just:  "Total"
- The Allocated Used - how much space is being used                                   "Used"
- The Total Free - how much space is available or unused                              "Free" 

which might look like: 

Pool Name  Total     Used         Free
  Pool 1        3TiB    287GiB     2.6TiB

drckeefe:
A detailed view could display (with metadata) similar to:
```
Pool Name            Total / Free      Data Allocated Total / Free         Metadata Allocated Total / Free   
  Pool 1             3TiB / 2.6TiB               384GiB / 106GiB                          512MiB / 507MiB
```
I wonder if we can leave either used or free out of the stats, as it is easy enough to calculate the other.  I believe I like to know
how much is available more often than how much was used.

mulkieran:
> In the output of all 5 stats a user might see (this is fictional)
> 
> ```
> Pool Name  Total Size  Allocated Size  Allocated Used  Allocated Free  Total Free
>   Pool 1      3TiB         384GiB        287GiB          106GiB         2.6TiB
> ```
> 
> This amount of information is good for a detailed view.
> 
> For a basic view I could see this:
> 
>     * The Total Size - the total amount of space in available to the pool, or just:  "Total"
> 
>     * The Allocated Used - how much space is being used,                                    "Used"
> 
>     * The Total Free - how much space is available or unused,                               "Free"
> 
> 
> which might look like:
> 
> ```
> Pool Name  Total     Used         Free
>   Pool 1   3TiB      287GiB     2.6TiB
> ```

@drckeefe The reason I like to keep the two "free" fields separate is that they mean two different things. The man pages explain those two different things, and also, how they are to be understood.

We could add them up into one "free" field, for summarizing. The two "free" regions do not overlap.

If we did that, we could stick to the Total/Used/Free triple for the summary view, which is helpful, and consistent with what we do with filesystems.

But the expanded view would be really necessary, on occasion, for the storage administrator or for troubleshooting.

mulkieran:
> A detailed view could display (with metadata) similar to:
> 
> ```
> Pool Name            Total / Free      Data Allocated Total / Free         Metadata Allocated Total / Free   
>   Pool 1             3TiB / 2.6TiB               384GiB / 106GiB                          512MiB / 507MiB
> ```
> 
> I wonder if we can leave either used or free out of the stats, as it is easy enough to calculate the other. I believe I like to know how much is available more often than how much was used.

I think it's helpful for the user to be able to tell things at a glance. I'm good at math, but bad at mental arithmetic, so it helps me to see all three values.

drckeefe:
>

mulkieran:
Closing this in favor of https://github.com/stratis-storage/stratis-cli/pull/865.

PR #865, opened 2022-03-09
Allow user to set filesystem limit on a pool

mulkieran:
Extends the CLI with a new command that allows the user to set the filesystem limit on a pool

* Does not make any change to the pool listing to show the filesystem limit. We anticipate that we may make a detailed pool view that will allow the user to see the filesystem limit and more detailed information about the pool usage.
* Does not make any change to the pool listing to show more detailed pool usage. We defer this and anticipate showing it in a detailed view. So, we stick to the Total/Used/Free view, where total is the total size of the all the data devices, used is the amount allocated to the mdv, and the two metadata devices + the amount actually used in the data device. Free is the difference between these two. I believe that the Free value overestimates by the relatively tiny amount of space required by Stratis metadata, but I'm willing to defer dealing with that problem until a later time. In any case, I believe this problem is much better dealt with on the stratisd side, and predates this PR by quite a bit.
* Adds additional tests as necessary.

### DIFF
--- a/docs/stratis.txt
+++ b/docs/stratis.txt
@@ -77,6 +77,9 @@ pool rebind keyring <pool_name> <keydesc>::
 pool unbind <(clevis|keyring)> <pool name> ::
      Unbind the devices in the specified pool from the specified encryption
      mechanism.
+pool set-fs-limit <pool name> <amount> ::
+     Set the limit on the number of file systems allowed per-pool. This number
+     may only be increased from its current value.
 pool explain <code> ::
      Explain any error code that might show up in the Alerts column when
      listing a pool.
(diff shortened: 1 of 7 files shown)


mulkieran:
Still draft, will need to remove the redirection commit.

drckeefe (approved):

jbaublitz (approved):

mulkieran:
rebased. Dropped CI retarget to PR branch.

mulkieran:
@bgurney-rh May I have your ack?

bgurney-rh (approved):

mulkieran:
rebased.

PR #882, opened 2022-05-05
Fix message for no alloc space

mulkieran:


bmr-cymru (approved):
New text makes the state of the pool clear and I think it's much better for users who might encounter this setting up pools.

mulkieran:
@jbaublitz CI hasn't been triggered, but it does all pass, so I'm not waiting on it.

PR #901, opened 2022-07-15
Expand block size

mulkieran:
Related https://github.com/stratis-storage/project/issues/50

Implements a new command: ```stratis pool extend-data``` and extends the output of ```blockdev list``` and ```pool list```. ```blockdev list``` shows the new size of the pool if different from the current in use-size. ```pool list``` displays an Alert on the pool, if different.

PR #906, opened 2022-08-29
Allow user to expand pool if block device is expanded

mulkieran:
Related https://github.com/stratis-storage/project/issues/50

Implements a new command: ```stratis pool extend-data``` and extends the output of ```blockdev list``` and ```pool list```. ```blockdev list``` shows the new size of the pool if different from the current in use-size. ```pool list``` displays an Alert on the pool, if different.

Supercedes #901

### DIFF
--- a/src/stratis_cli/_error_codes.py
+++ b/src/stratis_cli/_error_codes.py
@@ -120,12 +120,77 @@ class PoolAllocSpaceErrorCode(IntEnum):
         )
 
 
+class PoolDeviceSizeChangeCode(IntEnum):
+    """
+    Codes for identifying for a pool if a device that belongs to the pool has
+    been detected to have increased or reduced in size.
+    """
+
+    DEVICE_SIZE_INCREASED = 1
+    DEVICE_SIZE_DECREASED = 2
+
+    def __str__(self):
+        if self is PoolDeviceSizeChangeCode.DEVICE_SIZE_INCREASED:
+            return f"IDS{str(self.value).zfill(3)}"
+
+        if self is PoolDeviceSizeChangeCode.DEVICE_SIZE_DECREASED:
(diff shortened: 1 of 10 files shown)


mulkieran:
will fail stratisd tests.

mulkieran:
rebased.

mulkieran:
rebased.

drckeefe (approved):

src/stratis_cli/_error_codes.py
```diff
@@ -120,12 +120,76 @@ def from_str(code_str):
         )
 
 
+class PoolDeviceSizeChangeCode(IntEnum):
```
bmr-cymru, line 123: This is the only class in `_error_codes` that doesn't follow the `Pool.*ErrorCode` naming pattern (and also the only class to define an I(nfo) as well as W/E). PoolDeviceSizeChangeErrorCode is a bit of a mouthful at 29 chars. It stuck out a bit and I wondered if it might be confusing to come across the class name in isolation but I think I'm probably being nit-picky here :)

src/stratis_cli/_error_codes.py
```diff
@@ -120,12 +120,76 @@ def from_str(code_str):
...
+    """
+    Codes to extend if a device
```
bmr-cymru, line 125: Comment trails off?

bmr-cymru (changes requested):
I was a bit puzzled by the new class name in `_error_codes` but I think my initial expectation is probably looking for a foolish consistency :)

It would be nice to expand the class docstring comment though!

mulkieran:
Filed an issue for the naming question with error codes: https://github.com/stratis-storage/stratis-cli/issues/920

bmr-cymru (approved):

Issue #894, opened 2022-06-22
Add a facility for sending synthetic udev events


mulkieran:
The user running these will need root permissions. The value this adds is that the user need not know the details of the "/sys" directory.

mulkieran:
Related to https://github.com/stratis-storage/project/issues/50 since sending synthetic udev events might be what we need to trigger discovery of new device size.

mulkieran:
Tentatively re-assigning.

PR #911, opened 2022-09-14
Add debug subcommand for sending a synthetic uevent

bgurney-rh:
Related #894

### DIFF
--- a/src/stratis_cli/_actions/_debug.py
+++ b/src/stratis_cli/_actions/_debug.py
@@ -15,8 +15,10 @@
 Miscellaneous debugging actions.
 """
 
+# isort: STDLIB
+import os
 
-from .._errors import StratisCliEngineError
+from .._errors import StratisCliEngineError, StratisCliSynthUeventError
 from .._stratisd_constants import StratisdErrors
 from ._connection import get_object
 from ._constants import TOP_OBJECT
@@ -39,6 +41,26 @@ class TopDebugActions:  # pylint: disable=too-few-public-methods
         )
         if return_code != StratisdErrors.OK:  # pragma: no cover
             raise StratisCliEngineError(return_code, message)
+
+    @staticmethod
+    def send_uevent(namespace):
+        """
+        Issue a synthetic uevent from the CLI.
+        """
+        try:
+            realdevice = os.path.realpath(namespace.device, strict=True)
+            sysfs_device_uevent = (
+                realdevice.replace("/dev/", "/sys/class/block/") + "/uevent"
+            )  # pragma: no cover
+
+            with open(
+                sysfs_device_uevent, "w", encoding="utf-8"
+            ) as uevent_file:  # pragma: no cover
+                uevent_file.write("change")
+        except Exception as err:
+            raise StratisCliSynthUeventError(
+                f"Failed to send synth event: {err}"
+            ) from err
 
 
 class PoolDebugActions:  # pylint: disable=too-few-public-methods
--- a/src/stratis_cli/_error_reporting.py
+++ b/src/stratis_cli/_error_reporting.py
@@ -40,6 +40,7 @@ from ._errors import (
     StratisCliIncoherenceError,
     StratisCliParserError,
     StratisCliStratisdVersionError,
+    StratisCliSynthUeventError,
     StratisCliUnknownInterfaceError,
     StratisCliUserError,
 )
@@ -222,6 +223,12 @@ def _interpret_errors_1(
             f"that it would have been expected to execute as a result "
             f"of the command that you requested: {error}"
         )
+
+    if isinstance(error, StratisCliSynthUeventError):
+        return (
+            f"stratis reported an error in generating a synethetic "
+            f"udev event: {error}"
+        )  # pragma: no cover
 
     # Some method calls may have an assignable underlying cause.
     # At present, automated testing of any of these assignable causes is
--- a/src/stratis_cli/_errors.py
+++ b/src/stratis_cli/_errors.py
@@ -213,6 +213,12 @@ class StratisCliIncoherenceError(StratisCliRuntimeError):
     """
 
 
+class StratisCliSynthUeventError(StratisCliRuntimeError):
+    """
+    Raised if there was an error during a synthetic udev event generation.
+    """
+
+
 class StratisCliInUseError(StratisCliUserError):
     """
     Base class for if a request made of stratisd must result in a device being
--- a/src/stratis_cli/_parser/_debug.py
+++ b/src/stratis_cli/_parser/_debug.py
@@ -32,7 +32,15 @@ TOP_DEBUG_SUBCMDS = [
             help="Refresh all un-stopped pools.",
             func=TopDebugActions.refresh_state,
         ),
-    )
+    ),
+    (
+        "uevent",
+        dict(
+            help="Generate a synthetic uevent.",
+            args=[("device", dict(action="store", help="Path to device"))],
+            func=TopDebugActions.send_uevent,
+        ),
+    ),
 ]
 
 POOL_DEBUG_SUBCMDS = [
--- a/tests/whitebox/integration/test_debug.py
+++ b/tests/whitebox/integration/test_debug.py
@@ -15,18 +15,42 @@
 Test 'stratis debug'.
 """
 
+# isort: LOCAL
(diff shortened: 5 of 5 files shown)


src/stratis_cli/_actions/_debug.py
```diff
@@ -45,7 +45,12 @@ def send_uevent(_namespace):
         """
         Issue a synthetic uevent from the CLI.
         """
-        pass  # pylint: disable=unnecessary-pass
+        sysfs_device = _namespace.device.replace("/dev/", "/sys/class/block/")
```
bmr-cymru, line 48: I wonder if it's worth doing some simple validation checks on the device argument? It might help to give more clear error messages in the case that the user does something wrong.

E.g.:

 - Test that the passed device path exists and is a block device node
 - Verify the existence of the /sys/class/block/ directory and report an error

Another possible enhancement might be to check if the device argument is a symlink (for e.g. some `/dev/disk/by-*` link), and first try to resolve it with something like `os.path.realpath()`?

```python
>>> from os.path import realpath
>>> realpath("/dev/disk/by-uuid/a5c744e6-86d6-41de-ba87-f86bf4836242")
'/dev/nvme0n1p2'
>>> realpath("/dev/disk/by-path/pci-0000:02:00.0-nvme-1-part3")
'/dev/nvme0n1p3'
>>> 
```

bmr-cymru: If `realpath` cannot resolve a path in the file system it returns the string partly modified (up to the point where resolution failed - or the unmodified string if nothing could be found). This may not be desirable for our use, but this can be changed by passing `strict=True`:

```
>>> realpath("/a/disk/that/doesn't/exist")
"/a/disk/that/doesn't/exist"
>>> realpath("/a/disk/that/doesn't/exist", strict=True)
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.10/posixpath.py", line 395, in realpath
    path, ok = _joinrealpath(filename[:0], filename, strict, {})
  File "/usr/lib64/python3.10/posixpath.py", line 430, in _joinrealpath
    st = os.lstat(newpath)
FileNotFoundError: [Errno 2] No such file or directory: '/a'
```

bgurney-rh: OK, now I see.  When I try `sudo ./bin/stratis debug uevent /dev/falsepath/disk/by-path/pci-0000\:05\:00.0-part1`, I see this:

```
    with open(sysfs_device_uevent, "w", encoding="utf-8") as uevent_file:

FileNotFoundError: [Errno 2] No such file or directory: '/sys/class/block/falsepath/disk/by-path/pci-0000:05:00.0-part1/uevent'
```

So I tried this (in a commit that I'm about to push):
```
realdevice = os.path.realpath(_namespace.device, strict=True)

FileNotFoundError: [Errno 2] No such file or directory: '/dev/falsepath'
```

bmr-cymru: Gotcha - yes - in that case it's resolved the `/dev` but since that exists in the file system, and then left the trailing `falsepath` that it can't resolve to anything.

bgurney-rh:
I added `test_uevent()`, but I need to figure out the error that is resulting: `SystemExit: 2` (I believe this might be an ENOENT error, which would be 2; this is expected if the device is not in sysfs, which is most likely the case with the stratisd sim engine).

bgurney-rh:
Squashed into a single commit.

src/stratis_cli/_error_reporting.py
```diff
@@ -223,6 +224,12 @@ def _interpret_errors_1(
             f"of the command that you requested: {error}"
         )
 
+    if isinstance(error, StratisCliSynthUeventError):
+        return (
+            f"stratisd reported an error in generating a synethetic "
```
bmr-cymru, line 229: Maybe this would be better as something like "stratis encountered"? The daemon isn't involved in any of this so it could be confusing for users if the error mentions stratisd.

bgurney-rh: I changed it to "stratis reported", via a force-push of an amended commit.

src/stratis_cli/_actions/_debug.py
```diff
@@ -40,6 +42,25 @@ def refresh_state(_namespace):
         if return_code != StratisdErrors.OK:  # pragma: no cover
             raise StratisCliEngineError(return_code, message)
 
+    @staticmethod
+    def send_uevent(_namespace):
+        """
+        Issue a synthetic uevent from the CLI.
+        """
+        try:  # pragma: no cover
```
mulkieran, line 50: We do want to throw in one test. It should expect a failure with the exception type ```StratisCliSynthUeventError```. You'll want to add it in in stratis-cli/tests/whitebox/integration/test_debug.py. It's probably correct to make a new class which is a subclass of SimTestCase like the other class. Use examples that you'll find in lots of places that use ```self.check_error```; that's what you should be using here, because you will expect the error. Then try to move the ```no cover``` pragmas around so they allow only the minimal lines of code that will still allow 100% coverage.

mulkieran (changes requested):
I think the source looks reasonable; now time to work on testing.

bgurney-rh:
Rebased.  Preparing to add the new test.

src/stratis_cli/_actions/_debug.py
bgurney-rh: If I remove the `no cover` on both the "try" and "except" blocks, I get a coverage error on lines 52-58 (which would seem to be most, but not all, of the "try" block).

bgurney-rh:
Squashed into a single commit

src/stratis_cli/_actions/_debug.py
```diff
@@ -40,6 +42,25 @@ def refresh_state(_namespace):
...
+            realdevice = os.path.realpath(_namespace.device, strict=True)
+            sysfs_device_uevent = (
+                realdevice.replace("/dev/", "/sys/class/block/") + "/uevent"
+            )
+            uevent_str = "change"
```
mulkieran, line 55: Might as well drop the assignment here and just use the literal string ```"change"``` in the function call below.

mulkieran (changes requested):
I think these are my last requests...

src/stratis_cli/_error_reporting.py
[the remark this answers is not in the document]
bgurney-rh: Sorry; I must have lost this when I had to rebase.  Fixed now.

src/stratis_cli/_actions/_debug.py
```diff
@@ -40,6 +42,25 @@ def refresh_state(_namespace):
         if return_code != StratisdErrors.OK:  # pragma: no cover
             raise StratisCliEngineError(return_code, message)
 
+    @staticmethod
+    def send_uevent(_namespace):
+        """
+        Issue a synthetic uevent from the CLI.
+        """
+        try:  # pragma: no cover
+            realdevice = os.path.realpath(_namespace.device, strict=True)
```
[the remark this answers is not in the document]
bgurney-rh: If I do that, pylint returns `Undefined variable 'namespace'`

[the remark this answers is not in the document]
bgurney-rh: I see it now; I have to use `namespace` in the function declaration as well.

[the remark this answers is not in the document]
bgurney-rh: I'm getting into a mess where I keep seeing this change reverted, and running into merge conflicts with myself, so I'm going to force-push an update for all of the changes (which I'm planning to squash into a single commit anyway).

src/stratis_cli/_actions/_debug.py
bgurney-rh: Resolved via force-push

src/stratis_cli/_actions/_debug.py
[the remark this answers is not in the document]
bgurney-rh: Resolved via force-push

src/stratis_cli/_actions/_debug.py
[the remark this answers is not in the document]
bgurney-rh: I was able to get to a point where I only needed two "no cover" pragmas.  Resolved via force-push.

bgurney-rh:
Squashed into a single commit.

mulkieran (approved):

PR #929, opened 2022-10-24
Add support for starting pools by name

mulkieran:


mulkieran:
Expect stratisd based tests to fail.

mulkieran:
rebased.

mulkieran:
We should now expect all stratisd tests to succeed.

PR #951, opened 2023-01-23
Adapt stratisd-based tests to changes in stratisd repo

mulkieran:


mulkieran:
stratisd-based tests will fail now.

mulkieran:
Also need to pick up the glibc-static dependency

Issue #962, opened 2023-03-07
blockdev list -  displays the UUID of each device


drckeefe:
The list commands provide the uuid of each of the objects.  Would it be possible to add the uuids of each of the devices?

[root@sys ~]# stratis pool list
Name            Total / Used / Free    Properties                                   UUID   Alerts
p1     10 GiB / 1.06 GiB / 8.94 GiB   ~Ca,~Cr, Op   820e7e3e-b4cd-459c-a137-85b1876e7ef9   WS001 
[root@sys ~]# stratis fs list
Pool   Filesystem   Total / Used / Free             Created             Device                UUID                                
p1     fs1          1 TiB / 551 MiB / 1023.46 GiB   Feb 23 2023 18:25   /dev/stratis/p1/fs1   acd81b03-fe97-4188-904b-9453079d7f2c
[root@sys ~]# stratis blockdev list
Pool Name   Device Node   Physical Size   Tier
p1          /dev/vdb              5 GiB   DATA
p1          /dev/vdc              5 GiB   DATA

mulkieran:
It would be possible to list the UUIDs of the devices in the output of ```stratis blockdev list```

PR #972, opened 2023-03-29
Add UUID column to stratis blockdev list

bgurney-rh:
Related: #962


bgurney-rh:
This draft PR results in the following output:
```
Pool Name   Device Node   Physical Size   Tier   UUID                            
spool1      /dev/vdb1          8.00 GiB      0   34e634d872b146dc97aa9f80fbd393cb
```

Now I need to format the UUID.  I see `uuid_formatter()` used elsewhere; could we use it here?

mulkieran:
@bgurney-rh You should be able to imitate how we handle listing filesystem UUIDS in ```_logical.py```, by getting the correct uuid formatter from the namespace arguments.

bgurney-rh:
OK; I found the commit that added `get_uuid_formatter()`, in 560b07a3.

The output now looks like this:
```
Pool Name   Device Node   Physical Size   Tier   UUID                                
spool1      /dev/vdb1          8.00 GiB      0   34e634d8-72b1-46dc-97aa-9f80fbd393cb
```

...though I had to decide between `uuid_formatter` (in `_logical.py`) and `format_uuid` (in `_pool.py`).

I voted for `format_uuid` as it seemed more intuitive; should we change `_pool.py` to use that as well?

mulkieran:
@bgurney-rh The last test you should do is call blockdev list w/ the --unhyphenated-uuids option and verify that the UUIDs are now displayed unhyphenated.

After that, we can consider this PR ready.

bgurney-rh:
Unhyphenated:
```
./bin/stratis --unhyphenated-uuids blockdev list
Pool Name   Device Node   Physical Size   Tier   UUID                            
spool1      /dev/vdb1          8.00 GiB      0   34e634d872b146dc97aa9f80fbd393cb
```

Hyphenated:
```
./bin/stratis blockdev list
Pool Name   Device Node   Physical Size   Tier   UUID                                
spool1      /dev/vdb1          8.00 GiB      0   34e634d8-72b1-46dc-97aa-9f80fbd393cb
```

I added an extra commit for the variable rename.

drckeefe (approved):

mulkieran (changes requested):
@bgurney-rh Plz drop the changes to ```_pool.py``` and squash the commits, and we can go ahead and merge.

The way to think about the name difference is that since the names are both local to completely different functions it doesn't really matter if they are different. The other thing is that our pool listing code is far more heavyweight than our device and filesystem listing code, so that although that formatting function is obtained in the same way, it is actually used very differently, for example it is a field of the abstract base class ```List``` in ```_list_pools.py```.

bgurney-rh:
OK; I dropped the change to `_pool.py`, and squashed everything else into a single commit.

PR #981, opened 2023-05-01
Update introspect data for r6

mulkieran:


mulkieran:
All stratisd-based tests will fail.

PR #1026, opened 2023-09-25
Add typos target to Makefile

bgurney-rh:
Related: stratis-storage/project#663

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,6 +61,27 @@ jobs:
       - name: Run test
         run: ${{ matrix.task }}
 
+  check-typos:
+    runs-on: ubuntu-latest
+    container: fedora:38  # CURRENT DEVELOPMENT ENVIRONMENT
+    steps:
+      - name: Install dependencies
+        run: >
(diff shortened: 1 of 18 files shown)


.github/workflows/main.yml
```diff
@@ -61,6 +61,27 @@ jobs:
...
+        run: >
+          dnf install -y
+          clang
+          glibc-static
+          make
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          components: cargo
+          toolchain: 1.71.1  # LOWEST SUPPORTED RUST TOOLCHAIN
```
mulkieran, line 78: Use CURRENT instead of LOWEST. That's what I'll be using to run install typos, so that makes the most sense.

mulkieran (changes requested):
One request, otherwise looks good.

.github/workflows/main.yml
bgurney-rh: OK; amended to use 1.72.0.

mulkieran (approved):

Issue #1033, opened 2023-10-21
Should just use a function to get the dynamically generated classes that we import within methods


mulkieran:
This would be simpler and easier to understand.

PR #1041, opened 2023-11-27
Use dynamic build for Manager class

mulkieran:
Related #1033


mulkieran:
```make_dyn_class`` should just take a single key, indicating which class it is, currently it is too wordy.

PR #1042, opened 2023-11-27
Use dynamic build for Manager class

mulkieran:


mulkieran:
```make_dyn_class``` should memo-ize the dynamically generated class.

mulkieran:
The invocations of ```make_dyn_class``` are too wordy. They should be super simple, just one arg which very closely resembles the class name. ```make_dyn_class``` should then interpret that one argument.

mulkieran:
Need to add into a test that invokes CreatePool.

mulkieran:
This is not the correct way to do this. It should really be possible to make dynamic self-generating classes, and that would require working with dbus-python-client-gen.

mulkieran:
Obsoleted by the better idea of instantiating the whole class with only the methods needed for that invocation. Would still have to remember to add the absolute path check.

PR #1053, opened 2024-01-17
Implement Clevis-related parser errors as parser errors

### LINKED ISSUE #1037: Prefer a parser error to occur if option '--tang-url' is specified, but '--clevis=tang' is not.
Also, '--trust-url' and '--thumbprint' should require that '--tang-url' be set.

This is hard to enforce inside the parser using ```add_argument``` keyword arguments, but could be done at the very end of parsing, perhaps.

mulkieran:
Closes #1037

* Groups clevis-related options in the subcommand so they are grouped in the help text
* Some clevis-related options are conditionally dependent, i.e., they must be specified if the ```--clevis``` option is "tang" or "nbde", but otherwise not. Check if they are missing at parse time rather than at execution time so that, if they are missing, the error is understood to be a parse error (exit code 2).

### DIFF
--- a/src/stratis_cli/_actions/_utils.py
+++ b/src/stratis_cli/_actions/_utils.py
@@ -21,10 +21,6 @@ import json
 from uuid import UUID
 
 from .._constants import PoolIdType
-from .._errors import (
-    StratisCliMissingClevisTangURLError,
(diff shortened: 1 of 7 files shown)


mulkieran:
Moved tests to pure parser test class.

PR #1062, opened 2024-03-18
Clevis pin tidies

mulkieran:


mulkieran:
rebased

mulkieran:
Probably the solution to the problem of the incredibly messy Clevis configuration code, is to define a subclass of Action with a global object internally to store all the different Clevis stuff and then to define subclasses of that class for each possible option and then incrementally build that object as parser arguments are parsed.

mulkieran:
Should tidy up argument parsing first, probably.

mulkieran:
Blocked by https://github.com/stratis-storage/stratis-cli/pull/1073 .

PR #1073, opened 2024-04-22
Use Action subclass to parse clevis options

mulkieran:


bmr-cymru (approved):
My only reservation with this was the use of the hidden argparse option to trigger the post-parsing checks. I haven't seen that done before but it does clean up the logic in `get_info()` and `the_func()` nicely. Since argparse can't enforce the constraint for us this seems like a good way around it.

PR #1075, opened 2024-04-25
Add pool debug get-metadata command

mulkieran:
Related https://github.com/stratis-storage/project/issues/692

### DIFF
--- a/src/stratis_cli/_actions/_debug.py
+++ b/src/stratis_cli/_actions/_debug.py
@@ -16,6 +16,7 @@ Miscellaneous debugging actions.
 """
 
 # isort: STDLIB
+import json
 import os
 
 from .._errors import StratisCliEngineError, StratisCliSynthUeventError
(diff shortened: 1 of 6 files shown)


src/stratis_cli/_actions/_debug.py
```diff
@@ -91,6 +93,49 @@ def get_object_path(namespace):
...
+        )
+
+        if return_code != StratisdErrors.OK:  # pragma: no cover
+            raise StratisCliEngineError(return_code, message)
+
+        metadata = [int(x) for x in metadata]
+
+        if namespace.format is MetadataFormat.RAW:
+            print(metadata)
+            return
```
jbaublitz, line 127: What is the case you were thinking of when adding the raw option? Is it for if the JSON doesn't parse? I'm not entirely against it, but I think that it might be good to consider whether this will be used frequently.

jbaublitz (changes requested):
I just had one question about the raw format, but other than that, this looks good.

src/stratis_cli/_actions/_debug.py
mulkieran: @jbaublitz Since we're returning a string from stratisd, we can just drop the raw and have only two options, pretty or not.

jbaublitz (approved):

PR #1082, opened 2024-05-29
Python pydantic

mulkieran:


mulkieran:
I don't think the tests are complete enough to test the StoppedPool class, so a test success means nothing.

mulkieran:
We will have to postpone requiring pydantic for a while longer. Also, in order to use it, we would have to recursively convert any dbus-python property to appropriate native Python types before passing the objects to the pydantic parser. In a sense, we should have written the recursive parser long ago.

mulkieran:
Too many source code conflicts to be worth preserving.

PR #1091, opened 2024-08-05
Skip vetted tests

mulkieran:


bmr-cymru:
I wonder if we could use a custom decorator here and avoid the need to pass the test method name explicitly? I.e. using `__name__` from the decorator and then passing the result to `skipIf`? I haven't tried this out but it might allow us to avoid the need to pass the name and "skip requested" string explicitly. It's a small saving but if it works out it might be worth it.

mulkieran:
https://github.com/python/cpython/issues/122742#issuecomment-2272552249

PR #1098, opened 2024-09-13
Use pyproject.toml to specify global pylint allows

mulkieran:


mulkieran:
For pylint, there seems to be no way to add recursive configurations. So, you can really only put the global configuration in the pyproject.toml, and then do more local configurations via the command-line. For bandit, that technique doesn't work, all skips must be global.

mulkieran:
We are no longer using pylint, having switched to ruff.

PR #1100, opened 2024-09-25
Snapshot support

mulkieran:


mulkieran:
Expect test failures

mulkieran:
Expect all tests to pass

mulkieran:
Really this time...

bmr-cymru (approved):

mulkieran:
> Overall this looks good. The one thing I noticed in the PR is that we seem to create a new error for every error from the CLI when the D-Bus API returns an Identity result. Is this to specifically tie a good error message to each of these cases? Is there an advantage to not using the same error with different error messages as input? This is unrelated to the PR, but more something I noticed while reviewing this.

I think you're correct, that class inheritance tree looks unnecessarily bushy. We can prune it this autumn.

mulkieran:
Expecting tests to pass now...

PR #1127, opened 2025-02-05
Do not confuse pyright with overloaded str default

mulkieran:


mulkieran:
Note that if we were using typing module and could declare the type of ```interp``` we might be able to avoid both pylint and pyright errors.

Issue #1138, opened 2025-03-25
Adapt for new encryption commands, encrypt, unencrypt, re-encrypt


mulkieran:

mulkieran:
This was all accomplished in 3.8.2.

PR #1144, opened 2025-04-01
Post process idoptions in parser for filesystem list

mulkieran:
Related #1138


mulkieran:
Expect this to break...

mulkieran:
I've become convinced that this is not a good idea. I think our hidden options should probably only be used to do command-line parsing that simply can't be achieved via parsing instructions, e.g., value-dependent parsing, not other non-parsing things.

PR #1150, opened 2025-04-08
Use checkout/@v4 to check out stratisd repo

mulkieran:


mulkieran:
Once more it is succeeding!

PR #1156, opened 2025-04-29
Make ClevisInfo from Clevis command-line arguments in parser

mulkieran:


mulkieran:
Back to passing!

mulkieran:
This changes nothing, merely putting Clevis configuration calculation from command-line options in a single place, instead of several places. It removes a whole bunch of assertions by always calculating the Clevis configuration when parsing the command-line arguments and setting it in the parser namespace. Better than the, messier, alternatives.

PR #1157, opened 2025-05-01
Add a test for filesystem destroy when revert scheduled

mulkieran:


mulkieran:
The just-added test should fail.

mulkieran:
At this point, everything should pass...

PR #1167, opened 2025-06-04
Monkeytype

mulkieran:


mulkieran:
We do not need to annotate the source, we can invoke monkeytype just like python in Makefile targets.

PR #1169, opened 2025-06-04
Stop using LOWEST SUPPORTED ENVIRONMENT

mulkieran:
Related https://github.com/stratis-storage/stratisd/pull/3850


mulkieran:
Need to find a different way to install Python since the chosen way does not work on the Fedora containers.

PR #1172, opened 2025-06-17
Display alert if volume key missing or status unknown

mulkieran:
Related https://github.com/stratis-storage/stratisd/pull/3841

### DIFF
--- a/src/stratis_cli/_actions/_list_pool.py
+++ b/src/stratis_cli/_actions/_list_pool.py
@@ -19,7 +19,7 @@ Pool actions.
 import json
 import os
 from abc import ABC, abstractmethod
-from typing import List, Union
+from typing import List, Optional, Union
 
 # isort: THIRDPARTY
 from justbytes import Range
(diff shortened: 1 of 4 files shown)


mulkieran:
It is odd that we calculate the device size change codes separately.

src/stratis_cli/_actions/_list_pool.py
```diff
@@ -44,6 +50,24 @@
...
+        return None
+
+
+def _volume_key_loaded(mopool) -> Union[tuple[bool, bool], tuple[bool, str]]:
+    """
+    The string result is an error message indicating that the volume key
+    state is unknown.
+    """
+    result = mopool.VolumeKeyLoaded()
+    if isinstance(result, int):
```
jbaublitz, line 66: Does the type get unmarshalled as an integer? I would expect it to be of type `dbus.Bool` based on my interactions with the Python wrapper in other contexts.

mulkieran:
COPR is too busy to build right now. We shouldn't let that stop us.

jbaublitz (approved):

PR #1179, opened 2025-06-24
Use a list for the trigger key

mulkieran:


mulkieran:
Is Packit just really slow or does running instance still consider config file format invalid?

mulkieran:
Apparently, trigger key isn't supposed to accept a list...

PR #1200, opened 2025-08-13
Increase CURRENT DEVELOPMENT ENVIRONMENT to Fedora 42

mulkieran:
Related https://github.com/stratis-storage/project/issues/805


mulkieran:
The key is to add libatomic if using pyright for linting.

PR #1207, opened 2025-08-25
Support for encryption enhancements

mulkieran:
Related https://github.com/stratis-storage/stratisd/issues/3597

Supercedes https://github.com/stratis-storage/stratis-cli/pull/1186 

## Summary by CodeRabbit

* New Features
  * CLI: add pool encryption commands — on, off, reencrypt — with Clevis (Tang/NBDE), TPM2, and kernel keyring support; requires explicit --in-place for long-running in-place operations and prints confirmation when operations are initiated.
  * Pool listing: shows “Encryption Enabled: Last Time Reencrypted” when applicable.

* Documentation
  * Updated docs covering new commands, --in-place implications, and output fields.

* Tests
  * Expanded unit and integration tests for encryption flows, parsing, and long-running behavior.

* Chores
  * PR-triggered Copr builds now auto-start; build instrumentation coverage expanded.

PR #1213, opened 2025-09-05
Fix get_errors method

mulkieran:
Related #1207 

## Summary by CodeRabbit

* **Bug Fixes**
  * Error reports now follow only explicitly linked causes, reducing noise from unrelated context.
  * Broader exception handling improves detection and reporting of edge-case failures.
  * DBus service errors (including unknown-service cases) now yield a clear message indicating the management service isn’t running, aiding troubleshooting.


mulkieran:
So, the error in get_errors() was not quite benign, it was hiding an omission in ```_interpret_errors_0``` by following the context chain when the cause was missing has happens because ```NameHasNoOwner``` error is sometimes the context for the ```ServiceUnknown``` error.

PR #1219, opened 2025-09-19
Distinguish between different NoReply stack traces

mulkieran:
If stratis can't even get the stratisd version then it will never send any commands to stratisd. Let the user know this happened in that case.


jbaublitz (approved):

mulkieran:
I inserted a subtle bug...that I didn't test, one more try.

PR #1225, opened 2025-09-29
System file list

mulkieran:


mulkieran:
Note that all the tests are passing there is just a failure in coverage, not surprisingly.

mulkieran:
Closing this. The approach of an integrated filesystem display using both symlinks and D-Bus will be awkward and ugly. If we revise this, we can offer a debug option that will list filesystems based on symlinks data, only.

PR #1257, opened 2026-01-15
Support for in-place re-encryption, encryption, decryption

mulkieran:
## Summary by CodeRabbit

* **New Features**
  * Added pool encryption management: encrypt unencrypted pools, decrypt encrypted pools, and re-encrypt with a new master key
  * Pool information now displays the "Last Time Reencrypted" timestamp to track encryption operation history
  * Support for in-place encryption operations to enable long-running encryption transactions

* **Improvements**
  * Reduced D-Bus timeout from 120 to 60 seconds to improve responsiveness

### DIFF
--- a/docs/stratis.txt
+++ b/docs/stratis.txt
@@ -117,6 +117,25 @@ pool unbind <(clevis|keyring)> <pool name> [--token-slot <token slot>]::
      mechanism. MOVE NOTICE: The "unbind" subcommand can also be found under
      the "pool encryption" subcommand. The "pool unbind" subcommand that you
      are using now is deprecated and will be removed in stratis 3.10.0.
+pool encryption bind <(nbde|tang)> <(--uuid <uuid> |--name <name>)> <(--thumbprint <thp> | --trust-url)> <url>::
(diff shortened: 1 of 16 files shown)


mulkieran:
Right now, we do have the option of making the timeout shorter for our long-running operations than for our other operations. We would have to import a currently "private" method of dbus-python-client-gen. Then the user would not have to wait a full minute to get confirmation that the plan was initiated. We could build the methods as needed. In fact, this is probably what we should do for all methods.

mulkieran:
Long running operation decorator needs to specify the function that is allowed to run for a long time. Otherwise, if it's ```GetManagedObjects``` that is the long-running function, "Operation initiated" may be erroneously returned. Should allow it to take a list argument, just to be on the safe side.

jbaublitz:
Just one observation: I don't think GetManagedObjects every can be the long-running operation in the current state of stratisd but I'm aware that could change in the future with changes to our locking.

jbaublitz (approved):

mulkieran:
rebased, conflicts resolved.

mulkieran:
Need to catch a few more exceptions.

